PR #6, opened 2022-08-16
Dataset Changes

clarascu:
* Remove reused datasets from license and repository step
* Decouple DataRightsAndAccessControl and LegalRestrictions
* Display DataGeneration for new datasets only

fixes #79677


ZenoLC (approved):
good job!

PR #23, opened 2022-10-18
15 show search field for entries in drop down menu

lpandath:
I implemented the part of the frontend based on the queries I received as part of the output: https://github.com/tuwien-csd/damap-backend/issues/44

The user should be able to choose to search the ORCID database by name or ORCID ID or DMP list in the search box for their individual database.

I am including two images showing the queries I received in the search box.
 [Screenshot 2022-10-18 174515
](url)
Screenshot 2022-10-18 174551

### DIFF
--- a/libs/damap/src/lib/components/dmp/people/people.component.html
+++ b/libs/damap/src/lib/components/dmp/people/people.component.html
@@ -1,46 +1,69 @@
-<app-step-intro>{{'dmp.steps.people.intro' | translate}}
-</app-step-intro>
+<app-step-intro>{{ 'dmp.steps.people.intro' | translate }} </app-step-intro>
 
 <!-- CONTRIBUTORS -->
 <div [formGroup]="dmpForm">
   <ng-container formArrayName="contributors">
     <div *ngIf="contributors.length > 0">
-      <mat-card *ngFor="let contributor of contributors.controls; index as i" class="mat-selected">
+      <p>{{ 'dmp.steps.people.contributors' | translate }}:</p>
+      <mat-card
+        *ngFor="let contributor of contributors.controls; index as i"
+        class="mat-selected"
+      >
         <div class="mat-card-container">
           <div class="mat-card-left">
             <div>
               <mat-icon class="mat-icon-style">person</mat-icon>
-              {{contributor.value.firstName}} {{contributor.value.lastName}}
-              <mat-icon *ngIf="contributor.value.contact" class="mat-icon-contact" color="primary">
+              {{ contributor.value.firstName }} {{ contributor.value.lastName }}
+              <mat-icon
+                *ngIf="contributor.value.contact"
+                class="mat-icon-contact"
+                color="primary"
+              >
                 contact_mail
               </mat-icon>
-              <span class="sr-only">{{'dmp.steps.people.contact.person' | translate}}</span>
+              <span class="sr-only">{{
+                'dmp.steps.people.contact.person' | translate
+              }}</span>
             </div>
             <mat-icon class="mat-icon-style">mail_outline</mat-icon>
-            {{contributor.value.mbox}}
+            {{ contributor.value.mbox }}
             <app-orcid
-              *ngIf="contributor.value.personId?.identifier && (contributor.value.personId?.type === identifierType.ORCID)"
-              [orcidId]="contributor.value.personId?.identifier"></app-orcid>
+              *ngIf="
+                contributor.value.personId?.identifier &&
+                contributor.value.personId?.type === identifierType.ORCID
+              "
+              [orcidId]="contributor.value.personId?.identifier"
+            ></app-orcid>
           </div>
           <div [formGroupName]="i">
             <mat-form-field appearance="fill">
-              <mat-label>{{'dmp.steps.people.contributor.role' | translate}}</mat-label>
+              <mat-label>{{
+                'dmp.steps.people.contributor.role' | translate
+              }}</mat-label>
               <mat-select formControlName="role">
                 <mat-option
-                  *ngFor="let role of roles | keyvalue" [value]="role.key">
-                  {{translateEnumPrefix + role.value | translate}}
+                  *ngFor="let role of roles | keyvalue"
+                  [value]="role.key"
+                >
+                  {{ translateEnumPrefix + role.value | translate }}
                 </mat-option>
               </mat-select>
             </mat-form-field>
           </div>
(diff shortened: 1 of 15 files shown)


libs/damap/src/lib/components/dmp/people/people.component.ts
```diff
@@ -1,50 +1,66 @@
...
+  styleUrls: ['./people.component.css'],
 })
 export class PeopleComponent implements OnInit {
-
   @Input() projectMembers: Contributor[];
-
-  readonly roles: any = ContributorRole;
-  readonly identifierType = IdentifierType;
-
-  readonly translateEnumPrefix = 'enum.contributor.role.'
```
clarascu, line 24: Either keep that variable or replace it in the template with it's value.

libs/damap/src/lib/services/backend.service.ts
```diff
@@ -29,202 +37,220 @@ export class BackendService {
...
+        catchError(this.handleError('http.error.projectmembers'))
+      );
   }
 
+  getPersonSearchResult(searchTerm: string, serviceType: string): Observable<Contributor[]> {
+      return this.http
+        .get<Contributor[]>(
+          `${this.backendUrl}persons/search?q=${searchTerm}&searchService=${serviceType}`
+        )
+        .pipe(catchError(this.handleError('http.error.repositories.one')));
```
clarascu, line 149: Error message should be `http.error.person.search` not `http.error.repositories.one`.

clarascu (changes requested):
Just a few minor issues that would need to be fixed before merging (everything but the `mat-select` comment, that was just for reference), but other than that it looks good to me!

libs/damap/src/lib/components/dmp/people/people.module.ts
```diff
@@ -21,6 +23,7 @@ import {MatDialogModule} from '@angular/material/dialog';
     // Materials
     MatCardModule,
     MatSelectModule,
+    MatMenuModule,
```
clarascu, line 26: Do we still need that import?

libs/damap/src/lib/components/dmp/people/people.component.ts
lpandath: Not sure here, is not part of this issue. I think if you take this out will make issues on the html since seems to be used for adding the person role.

libs/damap/src/lib/components/dmp/people/people.module.ts
lpandath: Actually no. Tried another solution i forgot to take this out. Sorry for this.

libs/damap/src/lib/components/dmp/people/people.component.ts
clarascu: Yes, it's been removed, meaining roles aren't displayed correctly in the dropdown. So `translateEnumPrefix` should be kept :)

libs/damap/src/lib/components/dmp/people/people.component.html
```diff
@@ -52,50 +74,85 @@
...
+    "
+    (termToSearch)="searchContributor($event)"
+    (personToAdd)="addContributor($event)"
+  >
+  </app-person-search>
+  <div>
+    <mat-form-field appearance="fill">
+      <mat-label>{{ 'dmp.steps.people.search.service' | translate }}</mat-label>
+      <mat-select
+        (selectionChange)="onChangeService($event)"
```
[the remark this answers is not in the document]
clarascu: You can replace both, `[value]`and `(selectionChange)`, with `[(value)]` here, meaning you'll only need one variable :)

ZenoLC (changes requested):
backend PR GH-53 is ready for merging,
please use the list of services provided by it to compose the dropdown list for the person services.

lpandath:
@ZenoLC @clarascu branch is rebased with the next and tested within the solution of David's changes:: [de/44-enable-adding-contributors-sourced-from-orcid](https://github.com/rekt-hard/damap-backend/tree/de/44-enable-adding-contributors-sourced-from-orcid)

libs/damap/src/lib/components/dmp/people/people.component.html
```diff
@@ -1,46 +1,49 @@
-<app-step-intro>{{'dmp.steps.people.intro' | translate}}
-</app-step-intro>
+<app-step-intro>{{ 'dmp.steps.people.intro' | translate }} </app-step-intro>
 
 <!-- CONTRIBUTORS -->
 <div [formGroup]="dmpForm">
   <ng-container formArrayName="contributors">
     <div *ngIf="contributors.length > 0">
+      <p>{{ 'dmp.steps.people.contributors' | translate }}:</p>
```
clarascu, line 7: This line was removed in a previous PR, so please remove it here as well.

libs/damap/src/lib/components/dmp/people/people.component.ts
```diff
@@ -1,47 +1,61 @@
...
   searchResult$: Observable<Contributor[]>;
+  serviceConfigType$: ServiceConfig[];
 
-  constructor(private backendService: BackendService,
-              public dialog: MatDialog) {
-  }
+  constructor(
+    private backendService: BackendService,
+    public dialog: MatDialog,
+    public formService: FormService
```
clarascu, line 48: I think we can remove this as it doesn't seem to be used?

libs/damap/src/lib/components/dmp/people/people.component.ts
```diff
@@ -1,47 +1,61 @@
...
-
   @Input() projectMembers: Contributor[];
-
-  readonly roles: any = ContributorRole;
-  readonly identifierType = IdentifierType;
-
-  readonly translateEnumPrefix = 'enum.contributor.role.'
-
   @Input() dmpForm: UntypedFormGroup;
+  @Input() serviceConfigType: ServiceConfig;
```
clarascu, line 29: `@Input()` is not needed here as no value is passed from a parent component

clarascu (commented):
Just two minor things that need changes, but other than that it looks good to me!

libs/damap/src/lib/components/dmp/people/people.component.ts
clarascu: (For clarification: I'm talking about the `FormService`)

clarascu (approved):

ZenoLC (approved):

PR #27, opened 2022-10-21
setup default nginx conf

ZenoLC:


anussba (approved):
approved - looks good

PR #37, opened 2022-11-11
autosave scroll to the top of the new section

lpandath:


apps/damap-frontend/src/app/app.module.ts
```diff
@@ -59,6 +61,7 @@ export function HttpLoaderFactory(http: HttpClient): MultiTranslateHttpLoader {
 
     // Materials
     MatSnackBarModule,
+    MatStepperModule,
```
clarascu, line 64: This should not be needed here.

libs/damap/src/lib/components/dmp/dmp.component.ts
```diff
@@ -85,24 +100,45 @@ export class DmpComponent implements OnInit, OnDestroy {
...
+    ) as UntypedFormArray;
+    this.externalStorageInfo = this.dmpForm.get(
+      'externalStorageInfo'
+    ) as UntypedFormControl;
     this.repoStep = this.dmpForm.get('repositories') as UntypedFormArray;
     this.reuseStep = this.dmpForm.get('reuse') as UntypedFormGroup;
     this.costsStep = this.dmpForm.get('costs') as UntypedFormGroup;
   }
 
+  move(index: number) {
```
clarascu, line 114: This function doesn't seem to be used, should be save to remove it.

clarascu (changes requested):

clarascu (approved):

lpandath:
@clarascu i re-based with next and running i do not have any problem but Coverage fails in some lines also from before and im not sure why this happens. Maybe you can check that why this is failing? :/

clarascu (approved):

PR #38, opened 2022-11-14
[Bugfix] Make sure dmp-actions bar sticks to bottom

clarascu:
closes GH-31

### DIFF
--- a/libs/damap/src/lib/components/dmp/dmp.component.css
+++ b/libs/damap/src/lib/components/dmp/dmp.component.css
@@ -4,3 +4,8 @@
   bottom: 0;
   z-index: 10;
 }
+
+:host {
+  display: flex;
+  flex-direction: column;
+}


clarascu:
@lpandath Can you try this out and see if the issue still occurs? It's a bit hard to pin down.

lpandath:
> @lpandath Can you try this out and see if the issue still occurs? It's a bit hard to pin down.

So far, testing in Firefox or Chrome has worked fine, but what I could see upon investigation is that this action bar does not scale or behave responsively in the mobile version. 
image

clarascu:
> > @lpandath Can you try this out and see if the issue still occurs? It's a bit hard to pin down.
> 
> So far, testing in Firefox or Chrome has worked fine, but what I could see upon investigation is that this action bar does not scale or behave responsively in the mobile version. image

Yes, we currently disregard mobile view, since it's not a priority, but I'll keep it in mind for the future!

ZenoLC (approved):

PR #61, opened 2022-11-23
step 7 visual changes

lpandath:


libs/damap/src/lib/components/dmp/licenses/licenses.component.html
```diff
@@ -1,64 +1,73 @@
...
           </div>
-          <div class="mat-card-container">
+          <div id="licensingContainer" class="mat-card-container">
             <mat-form-field class="full-width">
-              <mat-label>{{'dmp.steps.licensing.question.license' | translate}}</mat-label>
-              <mat-select formControlName="license">
+              <mat-label>{{
+                'dmp.steps.licensing.question.license' | translate
+                }}</mat-label>
+              <mat-select [(value)]="defaultLicense">
```
clarascu, line 36: Using it like this sets the same license for all datasets in the view, but doesn't actually set the value for the DMP object. Setting a default license best happens in the [FormService](https://github.com/tuwien-csd/damap-frontend/blob/next/libs/damap/src/lib/services/form.service.ts#L286). `formControlName` then references this field in the form object. 
In `dev` mode (`nx serve ...`) the values stored in the form should be logged to the console and be updated every time any value changes, that way you can easily check what value is set and whether changes are reflected correctly.

libs/damap/src/lib/components/dmp/licenses/licenses.component.html
```diff
@@ -1,64 +1,73 @@
...
                 <mat-option *ngFor="let license of licenses" [value]="license.url | translate">
-                  {{license.name | translate}}
+                  {{ license.name | translate }}
                 </mat-option>
               </mat-select>
             </mat-form-field>
             <app-license-wizard (selectedLicense)="setLicenseSelectorResult($event, i)"></app-license-wizard>
           </div>
-          <div class="mat-card-container">
+          <div id="licensingContainer" class="mat-card-container">
```
clarascu, line 44: This `id` is used twice, maybe use  `class` instead?

libs/damap/src/lib/components/dmp/licenses/licenses.component.html
```diff
@@ -1,64 +1,73 @@
...
-                *ngIf="dmpForm.controls.project?.value?.end === dataset.value.startDate">
-                {{'dmp.steps.licensing.question.hint' | translate}}
+              <mat-hint *ngIf="
+                  dmpForm.controls.project?.value?.end === dataset.value.startDate
+                ">
+                {{ 'dmp.steps.licensing.question.hint' | translate }}
               </mat-hint>
             </mat-form-field>
           </div>
+          <div id="accessTypeContainer" class="mat-card-container">
```
clarascu, line 59: `dataAccess` disappears once it's been changed to something other than `OPEN`. Keep it visible, so that it may be changed again.

libs/damap/src/lib/components/dmp/data-deletion/data-deletion.component.html
```diff
@@ -3,33 +3,38 @@
...
           </mat-option>
         </mat-select>
       </mat-form-field>
       <mat-form-field>
-        <mat-label>{{'dmp.steps.data.deletion.question.date' | translate}}</mat-label>
-        <input matInput formControlName="dateOfDeletion" [matDatepicker]="picker">
+        <mat-label>{{
+          'dmp.steps.data.deletion.question.date' | translate
+          }}</mat-label>
+        <input matInput formControlName="dateOfDeletion" [matDatepicker]="picker" [(ngModel)]="planModel.end_time" />
```
clarascu, line 31: Using `[(ngModel)]` like this overwrites the form value everytime the form is loaded. Also `formControlName` and [(ngModel)] should not be used on the same form field since they represent two different ways of working with forms, but access the same field (see also: https://angular.io/guide/forms-overview). 
It's probably better to set the `dateOfDeletion` to the project end date when `delete` is changed (using the event emitted by toggle to distinguish user changes from programmatic changes).
(Can be done by patching the value of a given formgroup - it should look something like this and can be done in the `component.ts`: https://github.com/tuwien-csd/damap-frontend/blob/next/libs/damap/src/lib/services/form.service.ts#L291)

clarascu (changes requested):

libs/damap/src/lib/components/dmp/data-deletion/data-deletion.component.html
lpandath: Ill check this now.

clarascu (approved):

PR #63, opened 2022-11-23
change minor changes in step 10

lpandath:


clarascu (changes requested):
In this case not the font-size but the size (width) of the input field should be changed :)

Issue #65, opened 2022-11-25
Setup e2e tests


ZenoLC:
Setup e2e tests with cypress.
Integration with GitHub Actions currently not practicable, because we'll need a working backend + db and auth.
Additionally we'll need to define test cases that cover all main actions.

We can integrate e2e tests incrementally:

1. Run them manually before merging a PR. 
2. Include them in our own pipeline (CSD) and run them against a test backend.
3. Use them with GitHub actions.

###  Testcases:

- [ ] Create dmp
- [ ] Update dmp (for each section)
- [ ] Delete dmp
- [ ] Export dmp
- [ ] Create and view version
- [ ] Check dmp summary for complete forms
- [ ] Admin functionality

PR #67, opened 2022-11-29
Add e2e tests

clarascu:
## Description
Testing

#### What does this PR do?
Add e2e testing for the project

#### Breaking changes
No

#### Code review focus
Check if `nx e2e damap-frontend-e2e` runs without issue (needs a running backend).

### Checks
- [x] Documentation added
- [x] E2e tests created

closes GH-65


ZenoLC (changes requested):
missing change in package.json to adapt the e2e command to something runnable like e.g.`"ci:e2e": "nx run damap-frontend-e2e:e2e"`

also e2e test is still failing

ZenoLC:
now runs all green in headless mode, but still causes the following in watch mode:

```
Timed out retrying after 4000ms: Expected to find element: input#username, but never found it.

Because this error occurred during a before all hook we are skipping the remaining tests in the current suite: damap-frontend
```

ZenoLC (changes requested):
now runs all green in headless mode, but still causes the following in watch mode:
```
Timed out retrying after 4000ms: Expected to find element: input#username, but never found it.

Because this error occurred during a before all hook we are skipping the remaining tests in the current suite: damap-frontend
```

also add a command for watch mode so that users can more easily run it, e.g.
   ` "e2e": "npx nx run damap-frontend-e2e:e2e --watch=true",`

ZenoLC (approved):
good job

PR #82, opened 2023-01-04
Fix contributor filter pipe

clarascu:
## Description
Bugfix

#### What does this PR do?
Filter contributors from project members and search results based university id and person id.

#### Code review focus
Whether this now works as intended.

#### Dependencies
depends on changes from https://github.com/tuwien-csd/damap-backend/tree/zc/services-fix

### Checks


ZenoLC (approved):
very nice

PR #90, opened 2023-01-30
58 add export functionality for dmps with missing funder information

lpandath:
### Checks

- [ ] Summary updated
- [ ] Version view updated
- [ ] Documentation added
- [ ] Tests added
- [ ] E2e tests created

closes GH-
image

image
image


ZenoLC (changes requested):
Tests do not run successfully, see github actions pipeline.

libs/damap/src/lib/widgets/export-warning-dialog/export-warning-dialog.html
```diff
@@ -0,0 +1,29 @@
+<div>
+  <h1 mat-dialog-title>{{'dialog.export.title' | translate}}</h1>
+  <div mat-dialog-content>{{'dialog.export.content' | translate}}</div>
+  <div mat-dialog-actions>
+    <div class="container">
+      <ng-container>
+        <mat-form-field appearance="fill">
+          <mat-label>{{ 'dialog.export.type' | translate }}</mat-label>
+          <mat-select [(value)]="selectedTemplate">
```
clarascu, line 9: This should not be visible/selectable, if a funder is present.

libs/damap/src/lib/components/dmp/dmp-actions/dmp-actions.component.spec.ts
```diff
@@ -23,25 +27,28 @@ describe('DmpActionsComponent', () => {
...
         FormsModule,
         NoopAnimationsModule,
         TranslateTestingModule,
-        FormTestingModule
+        FormTestingModule,
+      ],
+      declarations: [
+        DmpActionsComponent,
+        SaveVersionDialogComponent,
+        ExportWarningDialogComponent,
```
clarascu, line 49: This should not be declared here, as is it imported by `ExportWarningModule`.

clarascu (changes requested):
The pop-up should not show the dropdown if a funder is provided.
The exports on the plans overview still uses the old API call. If funder information is available, use the original API call (template determined in the backend), if not, then show the dropdown (and use the new API which uses a template type).

Additionally we discovered an old bug: When selecting the JSON export, the pop-up is triggered - could you remove the pop-up entirely from this option?

apps/damap-frontend/src/environments/environment.ts
```diff
@@ -4,7 +4,7 @@
 
 export const environment = {
   production: false,
-  backendurl: 'http://localhost:8080/api/'
+  backendurl: 'http://localhost:8085/api/'
```
ZenoLC, line 7: you forgot to revert this :)

ZenoLC (changes requested):
there is now a new issue:
now the popup never contains the drop-down, even for projects without funder.
also when using the export function in the DMP view an error is thrown:
image
it will still export the file correctly, no idea what this new error is about.

PR #91, opened 2023-02-02
fetch and search projects on request via backend

rekt-hard:
refactor to use search result for contributors and projects

## Description

Feature/Refactoring

#### What does this PR do?

Use new search results for contributor and project search.

Projects are now fetched from the back end on demand. This allows for a more granular search experience and modification possibilities in the backend.
Before, projects were fetched once and the search was completely on the front end.

#### Breaking changes

Move project search from front end to back end.

#### Dependencies

depends on: https://github.com/tuwien-csd/damap-backend/pull/64

### Checks

- [ ] Summary updated
- [ ] Version view updated
- [ ] Documentation added
- [x] Tests added
- [ ] E2e tests created

related to https://github.com/tuwien-csd/damap-backend/issues/63


libs/damap/src/lib/components/dmp/project/project-list/project-list.component.html
```diff
@@ -1,13 +1,13 @@
-<p *ngIf="!selectedProject && projects?.length" translate>dmp.steps.project.select</p>
-<ng-container *ngIf="projects?.length else empty">
+<p *ngIf="!selectedProject" translate>dmp.steps.project.select</p>
+<ng-container *ngIf="!selectedProject">
```
clarascu, line 2: If `empty` is no longer used, you can remove `<ng-template #empty>` as well.

libs/damap/src/lib/components/dmp/dmp.component.html
```diff
@@ -1,7 +1,6 @@
 <mat-vertical-stepper #stepper [linear]="false" (selectionChange)="changeStep($event); changeStepPosition($event)">
   <mat-step label="{{'dmp.steps.project.label' | translate }}">
-    <app-dmp-project [projects]="projects$ | async"
-                     [loaded]="projectsLoaded$ | async"
+    <app-dmp-project [loaded]="projectsLoaded$ | async"
```
clarascu, line 3: `projectsLoaded$` is no longer used, so you can remove this.

clarascu (changes requested):
Since `projects` are no longer cached, you can remove them from the store altogether (State, Actions, Reducer & Selectors).

libs/damap/src/lib/components/dmp/dmp.component.ts
```diff
@@ -65,7 +59,6 @@ export class DmpComponent implements OnInit, OnDestroy {
   costsStep: UntypedFormGroup;
 
   // Resources
-  projectsLoaded$: Observable<LoadingState>;
   projects$: Observable<Project[]>;
```
clarascu, line 62: I think this might also no longer be needed. Could you check if that is safe to remove (and update line 249 ff accordingly)?

clarascu (approved):

Issue #85, opened 2023-01-25
Can not store costs after changing the cost value


rekt-hard:
**Describe the bug**
Issue with the costs section. After entering a value, it does not get parsed on the frontend but will be sent as a string to the backend.
The default entry has a value of 0. After entering something, it will be set to a string (like "7"). This can not be read by the backend.

**Is this a regression?**
Yes

**To Reproduce**
1. Create a new DMP
2. Add a dataset
3. Navigate to step 10 (Costs)
4. Add a new cost
5. Change the value of the cost to a valid number
6. Try to save the DMP
7. See error message

**Expected behavior**
Being able to define a custom cost entry with the correct cost value and store it with the DMP.

**Screenshots**
json without changing the cost value:
![image](https://user-images.githubusercontent.com/72449192/214571184-c46d5484-5bbc-4ec0-a35f-cb43a31fffa5.png)

json after changing the cost value
![image](https://user-images.githubusercontent.com/72449192/214571289-59d075cd-097c-4622-abab-9af10616f2ad.png)

**Environment**
 - Firefox
 - Damap version 2.30

**Additional context**

PR #93, opened 2023-02-07
update cost value form control

rekt-hard:
## Description

Bugfix

#### What does this PR do?

Subscribe to value changes for cost and transform the value to a number. This makes sure that further access of the form and its values will evaluate to a number and not a string.

### Checks

- [ ] Summary updated
- [ ] Version view updated
- [ ] Documentation added
- [ ] Tests added
- [ ] E2e tests created

closes GH-85


libs/damap/src/lib/services/form.service.ts
```diff
@@ -484,11 +484,22 @@ export class FormService {
   }
 
   private createCostFormGroup(): UntypedFormGroup {
+    let costFormControl = new FormControl(0, {
+      validators: currencyValidator(),
+      updateOn: 'change',
+    });
+    costFormControl.valueChanges.subscribe(n => {
```
clarascu, line 491: Try avoid using `.subscribe()` as `Subscriptions` exist until they are unsubscribed (unless it's a HTTP call). As this function will get called multiple times, we'll end up with mutliple (unused) subscriptions (see also: https://rxjs.dev/guide/observable#disposing-observable-executions).

clarascu (commented):
We weren't able to specify costs with decimals.

libs/damap/src/lib/services/form.service.ts
rekt-hard: Good point, thank you :)
In general, I am not sure if this is the right place to perform this conversion. It would be nice to have it here, so that further access does not have to make sure it is a number. On the other hand, it is not really used anywhere else. Which means, it would suffice to perform this conversion when sending the DMP to the back-end. Which will only happen, if the form is valid - thus there will always be a valid number in the text field (with or without decimals). What do you think?

ZenoLC:
was able to reproduce the cost error locally with the docker setup,
then found this link with the same issue https://stackoverflow.com/questions/74342327/jackson-parse-double-from-string-pure-number-format/74345893
and comparing the poms of public and custom project found out that the public project was running on quarkus 2.13, which imports jackson 2.13, but was also importing jackson-databind 2.14, which takes precedence over the one already contained in the quarkus packages as it has a shorter path in the dependency tree (as per maven pom rules). 
meaning: the bug should be solvable by updating the quarkus version to the newest one, and removing the explicit jackson package dependency.

PR #111, opened 2023-04-13
handle licenses as an enum for new datasets,

ZenoLC:
refs GH-https://github.com/tuwien-csd/damap-backend/issues/91

## Description

This PR changes the behavior of licenses in backend and frontend (has pendant backend PR).

#### What does this PR do?

The application will now exchange license data through enum values instead of URL (which was based on the maDMP standard).

#### Breaking changes

license information will be determined based on a new ID property, instead of url, so combining this frontend with older versions of the backend will break the license selection feature.

#### Dependencies

if tested, then must be done together with https://github.com/tuwien-csd/damap-backend/pull/94

### Checks

- [ ] Summary updated
- [ ] Version view updated
- [ ] Documentation added
- [ ] Tests added
- [ ] E2e tests created

closes GH-


libs/damap/src/lib/widgets/license-wizard/license-wizard-list.ts
```diff
@@ -275,6 +302,7 @@ export const bsd2c = {
   compatibility: ['bsd-2c', 'bsd-3c', 'apache-2', 'lgpl-2.1', 'lgpl-2.1+', 'lgpl-3', 'mpl-2', 'epl-1', 'cddl-1', 'gpl-2', 'gpl-2+', 'gpl-3', 'agpl-1', 'agpl-3']
 };
 export const apache2 = {
+  id: 'apache2',
   name: 'license.wizard.license.apache2.name',
   priority: 4,
   available: true,
```
rekt-hard, line 308: general question: does it make sense to have the licenses in the backend and in the frontend? Frontend should probably fetch available licenses, so it is easier to manage and updates only have to be performed in one place (single source of truth).

Other than that: depending on the name change of the enum values in the backend PR, this might need an update.

ZenoLC: yes it would make sense, but the current implementation has a very large feature fully implemented into the frontend, the license wizard, which is utilizing all the additional list properties in this file.
refactoring it into the backend would be a lot more effort tha it is worth.

rekt-hard: Understandable. Might be worth refactoring in a future issue

rekt-hard (approved):

rekt-hard (approved):

PR #128, opened 2023-05-30
Change template dropdown names to be more easy to read

lpandath:
## Description

This PR introduces an improvement to the user interface of the template selection dropdown for exports.

#### What does this PR do?
Currently, the template dropdown for choosing the template to export into lists the enum names as a String property. To improve readability and clarity, this PR updates the enum display names as follows:

    FWF: No change, stays as "FWF"
    SCIENCE_EUROPE: Changes to "Science Europe"

This change should make it easier for users to identify and select the correct export template.

### Checks

- [ ] Summary updated
- [ ] Version view updated
- [ ] Documentation added
- [ ] Tests added
- [ ] E2e tests created

closes GH-

### DIFF
--- a/libs/damap/src/lib/widgets/export-warning-dialog/export-warning-dialog.html
+++ b/libs/damap/src/lib/widgets/export-warning-dialog/export-warning-dialog.html
@@ -10,7 +10,7 @@
             <mat-option
               *ngFor="let id of dmpTemplate | keyvalue"
               [value]="id.key">
-              {{translateTemplatePrefixEnum + id.key | translate}}<br />
+              {{id.value | translate}}<br />
             </mat-option>
           </mat-select>
         </mat-form-field>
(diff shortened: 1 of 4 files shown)


libs/damap/src/lib/widgets/export-warning-dialog/export-warning-dialog.html
```diff
@@ -10,7 +10,7 @@ <h1 mat-dialog-title>{{'dialog.export.title' | translate}}</h1>
             <mat-option
               *ngFor="let id of dmpTemplate | keyvalue"
               [value]="id.key">
-              {{id.key}}<br />
+              {{translateTemplatePrefixEnum + id.key | translate}}<br />
```
rekt-hard, line 13: Do we want to translate the whole thing or just the `translateTemplatePrefixEnum`? Afaik, the organizations should not be translated.

rekt-hard, line 13: Should we use the `id.key` or the `id.value` for displaying?

rekt-hard (approved):

PR #132, opened 2023-07-06
[Bugfix] without project the form throws an error

ZenoLC:
## Description
Bugfix: Currently trying to save a DMP without a DMP connected to it, will cause the form to throw an error (trying to access project data with project null) and will cause the application to save an empty DMP. all subsequent save attempts will also create a new DMP , as the failure to load the DMP from the backend results in the ID also being empty. 

#### What does this PR do?

This fix will only query the funding data if the project data is also available. Thus resolving the error triggered when loading the form without project.

#### Breaking changes

No breaking changes.

#### Code review focus

Please check if there are other similar potential breaking points, or if the change has any repercussions further in the workflow

### Checks

- [ ] Summary updated
- [ ] Version view updated
- [ ] Documentation added
- [ ] Tests added
- [ ] E2e tests created

closes GH-

### DIFF
--- a/libs/damap/src/lib/services/form.service.ts
+++ b/libs/damap/src/lib/services/form.service.ts
@@ -112,7 +112,7 @@ export class FormService {
     this.form.patchValue({
       id: dmp.id,
       project: dmp.project,
-      funding: dmp.project.funding,
+      funding: dmp.project?.funding,
       data: {
         kind: dmp.dataKind,
         reusedKind: dmp.reusedDataKind,


rekt-hard (approved):
Nice, good job!

minor: how about also specifying the project as nullable in the [interface](https://github.com/tuwien-csd/damap-frontend/blob/fe9e9189bf8e4bc47ca4c68b29a03ff449a655f4/libs/damap/src/lib/domain/dmp.ts#L17)?

ZenoLC:
nevermind, the export does not work afte rthis fix, still open

ZenoLC:
Will be takebn care of by PR GH-133

Issue #138, opened 2023-09-14
Frontend build fails


SotosTsepe:
**Describe the bug**
Frontend build fails with `SyntaxError: Unexpected token '??='`.

**To Reproduce**
`docker compose up -d` from the damap-backend/docker directory

**Expected behavior**
Frontend build phase must exit without errors

**Screenshots**
![image](https://github.com/tuwien-csd/damap-frontend/assets/33498802/ec7ff57e-4cdc-426c-bbc1-346b8d5aa5b4)

**Environment**
 - Damap version [master]

PR #139, opened 2023-09-14
[hotfix] add nx version constraint to Dockerfile

SotosTsepe:
#### What does this PR do?
Constraints the nx version to <16.7.0 in the Dockerfile because the latest one is causing a build failure

closes GH-138


rekt-hard (approved):
Thanks!
Tested locally and works

PR #148, opened 2023-11-20
Fixed bug caused by chmod in Dockerfile.

ValentinFutterer:
Bug would cause different modules, like translate to stop working.

## Description

Bugfix

#### What does this PR do?

Removed the chmod in Dockerfile

closes GH-


rekt-hard (approved):
Great, thank you!

PR #162, opened 2024-01-24
Fix Bug: Reload Default List After Removing a Project

lpandath:
## Description

Displaying projects in the Project List component. Previously, when removing a project, the recommended projects were shown, but searching for new terms did not display search result

### Checks

- [ ] Summary updated
- [ ] Version view updated
- [ ] Documentation added
- [ ] Tests added
- [ ] E2e tests created

closes GH-


ValentinFutterer (changes requested):
The initial recommended project list (Recommended Project Numero Uno) is not loaded when I start creating a new DMP. It is only in the default list after it has been searched once and then deleted. As I understand it, the default list should always be visible, except when a project has been chosen.

rekt-hard (approved):
looking good!

ValentinFutterer (approved):

PR #191, opened 2024-04-02
config: hide data upload if FITS service not available

### LINKED ISSUE #59: disable FITS component/tab if FITS is not available (currently backend crashes)

larnhold:
## Description
Bugfix: Checks whether the backend offers a FITS service and if not hides the tab to upload files.

#### What does this PR do?
Previously the upload data tab was always displayed even when the backend did not offer that service. By checking the config sent by the BE the tab is now hidden if the service is not avialable.

#### Breaking changes
Requires the new field "fitsServiceAvailable" in "ConfigDO". If this field is not available then it assumes this service is not available and the "upload files" tab is hidden.

#### Code review focus
Previously only the "people" compoment accessed the config. In order to prevent the config to be loaded in each component seperately the rest call has been moved to the root component "dmp" and the config is passed down as an observable.

#### Dependencies
Depends on changes in the backend that add a new field to ConfigDO

### Checks

- [ ] Summary updated
- [ ] Version view updated
- [ ] Documentation added
- [ ] Tests added
- [ ] E2e tests created

closes tuwien-csd/damap-backend#59

### DIFF
--- a/libs/damap/src/lib/components/dmp/dmp.component.html
+++ b/libs/damap/src/lib/components/dmp/dmp.component.html
@@ -9,6 +9,7 @@
   </mat-step>
   <mat-step label="{{ 'dmp.steps.people.label' | translate }}">
     <app-dmp-people
+      [config$]="config$"
       [projectMembers]="projectMembers"
       [dmpForm]="dmpForm"
       (contactPerson)="changeContactPerson($event)"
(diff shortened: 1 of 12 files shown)


rekt-hard (commented):
Nice - working well!

In general: Should we manually unsubscribe from Observables to prevent memory leaks? Should not really matter in test cases but in the components themself.

larnhold:
Fair enough I will add that! Thank you for the review!

rekt-hard (approved):
Looking good, thank you!

We might want to introduce a base helper class that provides a list of subscriptions and takes care of unsubscribing from all subscriptions. So we can reuse logic in a few places. But can be shelved for now

PR #235, opened 2024-06-18
Shared: Make autocomplete panel responsive to resize of textarea-wrapper

### LINKED ISSUE #230: Global: Fix textarea-wrapper expansion bug
**Describe the bug**
The textarea can be expanded, but the popup with template text does not follow the expansion. Additionally, if a template text is clicked on, the whole text in the text area is overwritten. This can be undesired, since large amounts of text could be lost by accident. Change it so that the template text gets appended, in order to prevent users from losing progress.

image

**Is this a regression?**
No.

**To Reproduce**
Step 4 has a textarea-wrapper with template text.

**Expected behavior**
The popup should follow the expansion of the textfield; the text should not be overwritten by choosing a template text.

**Environment**

- Damap version 3.0.2

ValentinFutterer:
## Description

Bugfix

#### What does this PR do?

Before, the autocomplete panel would stay where it was, even if the textarea was resized - this can lead to loss of progress, since the autocomplete can overwrite the textarea. Now the panel resizes together with the textarea. I also removed the manual resizing option and made the textarea scale automatically up until 9 lines or ~ 200 words. If the limit is reached, the user has to scroll inside the textarea. According to Christiane almost no one writes that many words, so it should be fine this way. The manual resize button also took space away from the scrolling bar inside the textarea, that is another reason why i removed it, see the screenshot.

image

I also tried to remove the behaviour that the autocomplete option overwrites everything in the textarea, but it seems this will not be supported any time soon: https://github.com/angular/components/issues/12144
I also could not find a workaround :/

### Checks

- [ ] Summary updated
- [ ] Version view updated
- [ ] Documentation added
- [ ] Tests added
- [ ] E2e tests created

closes GH-230

### DIFF
--- a/libs/damap/src/lib/shared/textarea-wrapper/textarea-wrapper.component.css
+++ b/libs/damap/src/lib/shared/textarea-wrapper/textarea-wrapper.component.css
@@ -14,3 +14,7 @@ mat-form-field {
   );
   box-shadow: 0px 2px 5px rgba(0, 0, 0, 0.3);
 }
+
+.no-resize {
+  resize: none;
+}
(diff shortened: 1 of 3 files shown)


rekt-hard (approved):
good job

PR #248, opened 2024-06-28
git: update docker tag for release workflow

rekt-hard:
Use semantic version output for docker tag


lpandath (approved):
@rekt-hard  Looks good! Can you please rebase, and then you can merge it.

PR #249, opened 2024-07-03
Migrate theming and components to material 3

### LINKED ISSUE #242: Migrate theming and components from Angular Material 2 to Material 3
**Description**
Migrate our theming approach and components to Angular Material 3.

Theming Migration and Configuration: Update the existing theming based on Material 2 to the new theming system introduced in Material 3 and set based on the new design DAMAP the configuration needed.
Component update
Testing

**Describe alternatives you've considered**
[DAMAP Dashboard.pdf](https://github.com/user-attachments/files/15926810/DAMAP.Dashboard.pdf)

**Additional context**

_If this issue requests multiple changes consider using checkboxes or lists._

ValentinFutterer:
## Description

Refactoring

#### What does this PR do?

This PR introduces a change from the mat2 theming to the mat3 theming system.
I followed this guide closely: https://material.angular.io/guide/theming
The most changes concern the color system:
Mat3 introduced a new type of color palette, which i autogenerated from the old colors with the command `nx generate @angular/material:m3-theme` and `#006699` as primary and `#373737` as tertiary color input.

Then in `custom-theme.scss` i generated a light and dark theme from this newly generated color palette. The dark theme is unused, but i left it in for completions sake, if some universities want to use it. Could be removed though. The i included the light theme in every mat3 component in the whole application, which automatically styles every mat component in the correct color scheme.

In the next step, created global css variables in `styles.css` that use this new color system. I went through the whole application and replaced almost all hardcoded values with variables generated through the new color palette - which facilitates easy replacement of the whole color scheme. Some variables are still hardcoded, since the mat3 palettes dont include warning or succes, only god knows why.

I applied the variables through the whole application via closely following the material 3 color style guide: https://m3.material.io/styles/color/roles and choosing the correct colors acordingly. Also most colors slightly changed, but in most cases its not very noticeable or doesnt have a big impact on the overall look and feel.

The coloring of mat components has been completely reworked. Instead of using `color="primary"`, you have to create a custom css class, that overwrites internal color variables. So for each color on each component you would need to create a new css class. Granted, the theming colors most components automatically, but some still need to be changed manualls. For example, the raised buttons have a white background and colored text - to revert that you either have to use filled buttons or overwrite a variable. Also the process of finding variables is very exhausting - you need to use dev-tools to find them. There is an easier way that can skip the internal component variables, but it does not work for the mentioned button example and it only supports `primary`, `secondary` and `tertiary`, any other shades or surface colors you have to do with the internal variables. So I used the intended method only once. The custom css classes can be found in `custom-themes.css`.

Typography has been added to the theme - Roboto is the font of choice that is now automatically applied to all mat3 components. For all other components, i created a body font, that is relatively similar to the old one. If you want to use the predefined mat3 fonts, create a class like the body font. Here is a list of all predefined fonts and their properties: https://m3.material.io/styles/typography/type-scale-tokens

#### Breaking changes

New material3 palette + centralized color system are going to break institutional instances on a lower version.

#### Code review focus

Is the color system extendable?
Do you want something changed to make the transition for institutional instances easier?

### Checks

- [ ] Summary updated
- [ ] Version view updated
- [ ] Documentation added
- [ ] Tests added
- [ ] E2e tests created

closes GH-242


lpandath:
@ValentinFutterer After many approaches, this is how I was able to resolve the issues:
Added global focus styles to remove focus indicators for primary and on-primary buttons.
 Eliminated outline for active and focused mat-mdc-list-item.

apps/damap-frontend/src/app/components/layout/layout.component.css
```diff
@@ -7,9 +7,13 @@
   overflow: visible;
 }
 
+.mat-sidenav {
+  width: 301px;
+}
+
```
rekt-hard, line 13: do we want to set this to an absolute value?

ValentinFutterer: Yes, otherwise it scales with screen-width and it looks really bulky on bigger screens - at least thats my opinion.

lpandath (approved):

rekt-hard (approved):
massive work - thank you!

PR #273, opened 2024-07-29
Enhance person search: display email and ORCID in search results

### LINKED ISSUE #255: Add more info about person in search results
The name alone is often not enough to differentiate between people. Additional information like e-mail, ORCID should be shown in the result list.

lpandath:
## Description

Feature

#### What does this PR do?

Previously, only the first and last names were displayed in the search results. Now, additional information such as email addresses and ORCID identifiers are also shown. 

#### Breaking changes

This PR does not introduce any breaking changes.

#### Dependencies

This PR does not depend on or require any other changes.

### Checks

- [ ] Summary updated
- [ ] Version view updated
- [ ] Documentation added
- [ ] Tests added
- [ ] E2e tests created

closes GH-


libs/damap/src/lib/widgets/person-search/person-search.component.spec.ts
```diff
@@ -48,15 +48,19 @@ describe('PersonSearchComponent', () => {
...
     await input.enterText(mockContributor1.firstName);
-    const options = await input.getOptions({
-      text: mockContributor1.firstName + ' ' + mockContributor1.lastName,
-    });
 
-    expect(await input.getValue()).toBe(mockContributor1.firstName);
-    expect(options.length).toBe(1);
-    expect(await options[0].getText()).toBe(
+    const autocomplete = await loader.getHarness(MatAutocompleteHarness);
+    const options = await autocomplete.getOptions();
```
rekt-hard: The `input` variable can be reused instead of creating a new `autocomplete` variable

libs/damap/src/lib/widgets/person-search/person-search.component.html
```diff
@@ -12,7 +12,30 @@
...
-        {{ person.firstName }} {{ person.lastName }}
+        <div class="person-container">
+          <span class="person-name"
+            >{{ person.firstName }} {{ person.lastName }}</span
+          >
+          <div
+            *ngIf="
+              person.mbox ||
+              (person.personId?.identifier &&
+                person.personId?.type === identifierType.ORCID)
```
rekt-hard: Do we need the condition here? We check it afterwards per field so we could remove it here

libs/damap/src/lib/widgets/person-search/person-search.component.html
```diff
@@ -12,7 +12,30 @@
...
+            "
+            class="person-details">
+            <span *ngIf="person.mbox" class="person-info">{{
+              person.mbox
+            }}</span>
+            <span
+              *ngIf="
+                person.personId?.identifier &&
+                person.personId?.type === identifierType.ORCID
```
rekt-hard: It may be any identifier so just checking for ORCID is a bit restrictive. If there is an identifier, it would be good to just display it as: `<identifier type>:<identifier value>`. So this check and the import in the `.ts` file can be removed.

rekt-hard (commented):
Thank you - already much easier to identify a person with the additional info.
The details overflow a bit
Screenshot from 2024-07-29 17-51-48

Can we make the result list wider?

libs/damap/src/lib/widgets/person-search/person-search.component.html
lpandath: Yes we need it, in case one of them is missing, i do not want to show an empty space instead.

libs/damap/src/lib/widgets/person-search/person-search.component.html
lpandath: Yes sure we can change that, as in the issue itself was not mention, this was my approach.

libs/damap/src/lib/widgets/person-search/person-search.component.html
rekt-hard: good point, makes sense

rekt-hard (commented):
Looking good.

Is it possible to make the person result list as wide as the project result list? Just to be consistent and catching cases, where people have `really` long e-mail addresses.

project:
Screenshot from 2024-07-30 16-00-05


person:
image

lpandath:
> Looking good.
> 
> Is it possible to make the person result list as wide as the project result list? Just to be consistent and catching cases, where people have `really` long e-mail addresses.
> 
> project: Screenshot from 2024-07-30 16-00-05
> 
> person: image

Based on the new mockups the content will change, so i made this as clear as possible, so that when we change the content between the right and left is easier to make the future changes. 
image
image

rekt-hard:
I think we could use the same components we use when displaying projects (mat-selection-list, mat-list-option (style with class `matListItemTitle` and `matListItemLine`)) so they look the same. May have some resources next week to give it a try

rekt-hard:
Have not updated the test cases and conditional display of mail/identifier is missing. Will update if we agree on this design.
Would also move the service dropdown into the person search component, so it is easier to style the result list beneath the form fields.

Current design with using the mat list display options (same as project result list):
image

ValentinFutterer:
I agree with David here, the styling using mat-list-item makes a lot of sense imo. Something to add from my side: The icons are scaled with font-size, which just scales the icon, but leaves the box around untouched. This leads to wrong alignment and way too much space between the two lines. The icon scaling should be done using scale/transform, the amount of scaling should be proprtional to the font-size in the same line. (One could use the body-typography variable for this purpose). Alternatively, the scaling could still be done using font, but then the width/height has to be scaled too.

rekt-hard:
In general, there is an issue with all icons that use the `mat-icon-style` class. As you mentioned, the width and height of the container will not adapt by changing the font size.

Tried quite a few things now. When scaling, the width/height stays the same and the distance from the container to other elements will not change. It is a purely cosmetic change, since width/height will be dictated by other rules.

We could:
1. Use a more specifc css rule, that will overwrite the height and width and then set the font-size. Since there are some other elements with a hard coded size (e.g. orcid element) it might introduce some UI changes - mostly unaligned icons and text.
```css
.mat-icon.mat-icon-style : {
  width: auto;
  height: auto;
  font-size: smaller;
}
```

2. Use the `inline` attribute of the mat-icon element to automatically fit the parent container. For the current use case, this would be enough (see Screenshot). We could go ahead and change that for other icons as well (e.g. project list), but have to check if no other issues are introduced by doing so. Definitely a future task IMO.
image

lpandath:
> In general, there is an issue with all icons that use the `mat-icon-style` class. As you mentioned, the width and height of the container will not adapt by changing the font size.
> 
> Tried quite a few things now. When scaling, the width/height stays the same and the distance from the container to other elements will not change. It is a purely cosmetic change, since width/height will be dictated by other rules.
> 
> We could:
> 
>     1. Use a more specifc css rule, that will overwrite the height and width and then set the font-size. Since there are some other elements with a hard coded size (e.g. orcid element) it might introduce some UI changes - mostly unaligned icons and text.
> 
> 
> ```css
> .mat-icon.mat-icon-style : {
>   width: auto;
>   height: auto;
>   font-size: smaller;
> }
> ```
> 
>     2. Use the `inline` attribute of the mat-icon element to automatically fit the parent container. For the current use case, this would be enough (see Screenshot). We could go ahead and change that for other icons as well (e.g. project list), but have to check if no other issues are introduced by doing so. Definitely a future task IMO.
>        image

Are icons new in place? Looking good!

I will review the current state and see the changes to be done.

libs/damap/src/lib/widgets/person-search/person-search.component.html
```diff
@@ -8,33 +8,29 @@
       (input)="search(searchBox.value)" />
     <mat-icon matSuffix>search</mat-icon>
   </mat-form-field>
-  <div
-    *ngIf="currentSearchTerm && result.length > 0"
```
lpandath: @rekt-hard The reason I left out this condition is because we don't want to show the list if the search box is empty. Is there a reason why this was removed?

rekt-hard: My bad. In the projects section, we show recommended projects if the input is empty. Will revert it

ValentinFutterer:
I am happy with the inline solution, though we should create an issue to fix the icon alignment globally, just as you said. Could you add a small indent to the email and id, just like it was before the icons where added:

image

rekt-hard:
Will try to align it like the project result list

lpandath:
> I am happy with the inline solution, though we should create an issue to fix the icon alignment globally, just as you said. Could you add a small indent to the email and id, just like it was before the icons where added:
> 
>
@ValentinFutterer As discussed we can change that after the main changes are in if needed. 

PR is updated.

ValentinFutterer:
Yes from my side it looks ready to merge.

ValentinFutterer (approved):

PR #285, opened 2024-08-07
Bug fix: Source field of Reused data can be updated

### LINKED ISSUE #283: Reused data source added manually cannot be updated
**Describe the bug**
The source field of manually entered resued data sources cannot be updated (it reverts to its original value, works fine for the dataset title and the description).

**Is this a regression?**
Yes/No.

**To Reproduce**

1. Create a new manually added "Reused data"
2. Enter sample values for the 3 fiels
3. Modify the values with the pencil button
4. Press the update button, the title and description are updated but not the source

**Expected behavior**
The new value for the source should be the one provided in the update form.

**Screenshots**
image

**Environment**

- Damap version [e.g. 2.1.0]

**Additional context**

GeoffreyKarnbach:
## Description

Bugfix

#### What does this PR do?

The field "Source" in the "Reused Data" section can be updated again. Before, the updated value did not get persisted.
In the constructor, a deep copy of the datasetId object was made, as objects from the store are by default read-only.

### Checks

- [ ] Summary updated
- [ ] Version view updated
- [ ] Documentation added
- [ ] Tests added
- [ ] E2e tests created

closes GH-283

### DIFF
--- a/libs/damap/src/lib/components/dmp/specify-data/dataset-dialog/dataset-dialog.component.ts
+++ b/libs/damap/src/lib/components/dmp/specify-data/dataset-dialog/dataset-dialog.component.ts
@@ -47,7 +47,7 @@ export class DatasetDialogComponent {
     this.dataset.patchValue(data.dataset);
     this.mode = data.mode ?? this.mode;
     if (data.dataset.datasetId) {
-      this.datasetId = data.dataset.datasetId;
+      this.datasetId = { ...data.dataset.datasetId };
     }
   }
 


libs/damap/src/lib/components/dmp/specify-data/dataset-dialog/dataset-dialog.component.ts
ValentinFutterer, line 1: Hey your solution works, but it is not solving the root of the issue. Here the (immutable) data is passed inside the dialog: https://github.com/tuwien-csd/damap-frontend/blob/cae32d40f1b40095b6c75da4039a4e9680bb453b/libs/damap/src/lib/components/dmp/specify-data/dataset-dialog/dataset-dialog.component.ts#L49-L51 It would be preferable to create a copy here, instead of everytime a change occurs. (This also removes the need for getters and setters)

GeoffreyKarnbach: Implemented in the last commit

ValentinFutterer (approved):
Nice, bug is fixed.

PR #287, opened 2024-08-12
Global: Updated step 8 retention component

### LINKED ISSUE #48: Step 8 changes
- move "Set retention period for datasets" component in between list of selected repositories and list of selectable repos.
- set retention period to 10 years by default
- filter list of available datasets here to the list of actually assigned datasets (assigned to repos)
- Add text to retention period collapsible component (ask RDM for text resource).

GeoffreyKarnbach:
## Description

Feature

#### What does this PR do?

Move Retention Period component in between list of selected repositories and list of selectable repos
Filtered to only show datasets in the component that have at least one repository assigned
Added tooltip banner with further information

### Checks

- [ ] Summary updated
- [ ] Version view updated
- [ ] Documentation added
- [ ] Tests added
- [ ] E2e tests created

closes GH-48

### DIFF
--- a/libs/damap/src/lib/components/dmp/repo/retention-period/retention-period.component.html
+++ b/libs/damap/src/lib/components/dmp/repo/retention-period/retention-period.component.html
@@ -7,24 +7,36 @@
         </mat-panel-title>
         <mat-icon class="color-icon">edit_calendar</mat-icon>
       </mat-expansion-panel-header>
-      <ng-container *ngFor="let dataset of datasets.controls; index as i">
-        <ng-container *ngIf="dataset.value.source === datasetSource.NEW">
-          <div formArrayName="datasets">
-            {{ dataset.value.title }}
-            <mat-form-field appearance="fill" [formGroupName]="i">
-              <mat-label>{{
-                "dmp.steps.retentionPeriod.question.retentionPeriod" | translate
-              }}</mat-label>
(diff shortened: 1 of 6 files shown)


libs/damap/src/lib/components/dmp/repo/retention-period/retention-period.component.html
```diff
@@ -22,6 +26,9 @@
                 </mat-option>
               </mat-select>
             </mat-form-field>
+            <button mat-stroked-button class="unclickable">
```
ValentinFutterer: I would avoid using buttons as textboxes - this can be confusing, especially for users with disabilities. You could intead just keep blank text with some styling, or maybe create mat-cards, with the name as title. if the name is too long, you could truncate and show the rest via tooltip.

ValentinFutterer (commented):
Using the same dataset in two different repos causes this: 
image
image

Also, could you add a little padding here, thank you.

image

libs/damap/src/lib/components/dmp/repo/retention-period/retention-period.component.html
GeoffreyKarnbach: Changed to using span tags, styled in a similar fashion:

image

ValentinFutterer (approved):
Nice, ready to merge

PR #290, opened 2024-08-12
Global: Prevent dupplicate DOI for reused data

### LINKED ISSUE #254: Check duplicates on dataset via Identifier
Don't allow the same DOI to be added twice in the dataset list.

GeoffreyKarnbach:
## Description

Feature

#### What does this PR do?

Added frontend validation, check that a new added DOI is not in the datasets array, if it is, no search is run, and an error message is displayed.

#### Dependencies

Related Backend Issue:

https://github.com/tuwien-csd/damap-backend/issues/239

### Checks

- [ ] Summary updated
- [ ] Version view updated
- [ ] Documentation added
- [ ] Tests added
- [ ] E2e tests created

closes GH-254

### DIFF
--- a/libs/damap/src/assets/i18n/templates/en.json
+++ b/libs/damap/src/assets/i18n/templates/en.json
@@ -426,7 +426,8 @@
       "maxlength": "Value must not exceed {{value}} characters.",
       "pattern.currency": "Invalid input, please use the following format: 1234.56",
       "orcid": "Invalid ORCID iD.",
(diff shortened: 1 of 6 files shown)


rekt-hard (commented):
works great.

any reason to call it `dupplicate` instead of `duplicate` or just a typo?
could you add a test case to cover this functionality? Simply searching for the same dataset twice should do the trick.

rekt-hard (approved):

PR #294, opened 2024-08-19
Bugfix: Log out user if token experied

### LINKED ISSUE #279: Access token refresh fails when device is in sleep mode
**Describe the bug**
When the device is in sleep mode, the access token is not refreshed as it should be. This leads to an 401 error - for the user it looks like the page suddenly just started not responding. The workaround is to open a new tab with the same url - this is of course not apparent for users.

**Is this a regression?**
No.

**To Reproduce**

keep the device in sleep mode, when the refresh is scheduled. The you can observe the behaviour.

**Expected behavior**
the application should either refresh the token when the user comes back online, or should log the user out. You can take inspiration from here https://github.com/manfredsteyer/angular-oauth2-oidc/issues/820.

**Environment**

- Chrome 127.0.6533.89
- Damap version 4.0.0

GeoffreyKarnbach:
## Description

Bugfix

#### What does this PR do?

If a device was in sleep mode, the auth token did not refresh. Now, before any HTTP request, we use an Angular Interceptor and check if the token is still valid (or is about to expire). In this case, we display an alert and logout the user.

### Checks

- [ ] Summary updated
- [ ] Version view updated
- [ ] Documentation added
- [ ] Tests added
- [ ] E2e tests created

closes GH-279


libs/damap/src/lib/interceptors/auth.interceptor.ts
```diff
@@ -0,0 +1,41 @@
...
+    const accessToken = this.oAuthService.getAccessToken();
+    if (accessToken == null) {
+      return next.handle(req);
+    }
+
+    const date = new Date(this.oAuthService.getAccessTokenExpiration());
+    const timeLeft = date.getTime() - new Date().getTime();
+
+    if (timeLeft < 5) {
+      alert('Your session has expired. Please log in again.');
```
rekt-hard: Can we use the feedback service here?

libs/damap/src/lib/interceptors/auth.interceptor.ts
```diff
@@ -0,0 +1,41 @@
...
+      this.authService.logout();
```
rekt-hard: suggestion: would it make sense to let the user try to login again? A user might have entered some data which would be lost by simply logging out.

libs/damap/src/lib/interceptors/auth.interceptor.ts
```diff
@@ -0,0 +1,41 @@
...
+      return new Observable<HttpEvent<any>>();
```
rekt-hard: suggestion: try to redirect to the current page. At least a user will end up at the same place after logging in again.

libs/damap/src/lib/interceptors/auth.interceptor.ts
```diff
@@ -0,0 +1,41 @@
...
+    req: HttpRequest<any>,
+    next: HttpHandler,
+  ): Observable<HttpEvent<any>> {
+    const accessToken = this.oAuthService.getAccessToken();
+    if (accessToken == null) {
+      return next.handle(req);
+    }
+
+    const date = new Date(this.oAuthService.getAccessTokenExpiration());
+    const timeLeft = date.getTime() - new Date().getTime();
```
rekt-hard, line 31: `AuthService` has a method called `isAuthenticated` . Would this also perform an expiration check?

rekt-hard (commented):
In general, it probably makes sense to implement such knowledge in `AuthService`. This should be the place for customizations as well. The more often we directly access the `OAuthService` in different components, the more extra work and care has to be put into customizations.

`OAuthService` provides some listenable events. 
```    /**
     * Informs about events, like token_received or token_expires.
     * See the string enum EventType for a full list of event types.
     */
    events: Observable<OAuthEvent>;
```
Instead of checking for expiration on every single request, is it possible to only listen to certain events?

@ValentinFutterer @lpandath  what is your opinion on this?

libs/damap/src/lib/interceptors/auth.interceptor.ts
GeoffreyKarnbach: I tried it didn't work as it was not in the same structure if I remember right (libs vs apps folder)

libs/damap/src/lib/interceptors/auth.interceptor.ts
GeoffreyKarnbach: isAuthenticated did not work, I tried with it

libs/damap/src/lib/interceptors/auth.interceptor.ts
ValentinFutterer: Due to the redirect when logging in, the data would be lost anyway, or am i missing something.

ValentinFutterer:
> Instead of checking for expiration on every single request, is it possible to only listen to certain events?
> 
> @ValentinFutterer @lpandath what is your opinion on this?

In principal yes, I am with you. But the issue is, that these events will be lost when in sleep mode. Otherwise the automatic token refreshing would prevent this error in the first place.

lpandath:
> > Instead of checking for expiration on every single request, is it possible to only listen to certain events?
> > @ValentinFutterer @lpandath what is your opinion on this?
> 
> In principal yes, I am with you. But the issue is, that these events will be lost when in sleep mode. Otherwise the automatic token refreshing would prevent this error in the first place.

Listening to events like token_received or token_expires instead of checking the token expiration on every request sounds like a good idea. But as mentioned, there’s a risk of missing these events if the device goes into sleep mode. In that case, maybe the session might expire without us knowing, which could cause issues... hmm, also isAuthenticated(), it would be great if this method could also check whether the token has expired.

libs/damap/src/lib/interceptors/auth.interceptor.ts
rekt-hard: I was thinking about the following:

There is a function called `initLoginFlowInPopup`in the `OAuthService`. We could allow the user to re-login within the popup:
- If successful, continue with the request as user is logged in again.
- If not successful, show popup again
- If user closes popup manually, logout and redirect

This assumes that a manual close can be recognized (to be investigated). We have to look at the specifics of the function to get a better picture of what is possible.

ValentinFutterer:
Looks pretty good to me and i cannot think of another solution than to check at every usage of a token, if its still valid. Sleep mode throws a big wrench into everything. Your solution also covers the case of the refresh token expiring - so thats a nice bonus.

The only thing I would suggest, is to first try a refresh with the refresh token when the access token is expired. If this refresh fails, it means that most likely the refresh token is expired and we should start the login flow again. This way we can prevent unnecessary logins.

@kohlrabicode what do you make of this solution?

PR #302, opened 2024-08-29
Improve sidenav footer behavior

### LINKED ISSUE #299: Add footer with mobile responsiveness
**Changes**
Enhance the side navigation -  footer: showing icons in the mobile view and text in the desktop view.

**Additional notes**
List any notes you collected that support the original task and include a screenshot if applicable.

**Assumptions**
List any assumptions that you are aware of but still need to be reviewed.

**Acceptance Criteria**
Update the side navigation to display only icons when viewed on mobile or small devices, while retaining the full text in the desktop view.
Add a responsive footer to provide useful links or information across all devices.

#### Icon update
Due to an update in the icon set, the usage of outline icons has to be changed. For details see here: https://github.com/tuwien-csd/damap-frontend/pull/302#pullrequestreview-2275869757
TLDR;
`<mat-icon> icon_name_outline</mat-icon>`
should be changed to
`<mat-icon class="outline">icon_name</mat-icon>`

**Risks? (optional)**
Potential risks and how to best mitigate them.

lpandath:
#### Code review focus

Improved sidenav footer to show a responsive icon when collapsed.
Updated background color based on new mockups; further adjustments will be made as the responsive layout is finalized.

_Note_: The icon can be changed if needed.

### Checks

- [ ] Summary updated
- [ ] Version view updated
- [ ] Documentation added
- [ ] Tests added
- [ ] E2e tests created

closes GH-


apps/damap-frontend/src/app/components/layout/layout.component.html
```diff
@@ -70,24 +69,55 @@
...
+      </div>
+      <div class="version-information">
+        <a [routerLink]="'/gdpr'" rel="noopener">{{
+          "layout.gdpr" | translate
+        }}</a>
+        <span class="version-text"
+          >{{ "layout.version" | translate }}: {{ version }}</span
+        >
+      </div>
+    </div>
```
rekt-hard, line 119: Why should it not be possible to access these links, if the menu is collapsed or on a small screen? Not an expert, but there might be legal issues as a user does not have access to important documents.

So some way or another, these documents must still be accessible through the UI

apps/damap-frontend/src/app/components/layout/layout.component.html
```diff
@@ -70,24 +69,55 @@
...
+    <div *ngIf="isCollapsed" class="footer-icon">
+      <mat-icon class="menu-icon-size">admin_panel_settings</mat-icon>
```
rekt-hard: Right now, only the icon is displayed. Nothing happens when clicking on it. Should there be some interaction possibility?

rekt-hard (commented):
I think it makes more sense to have a fixed side menu/navbar taking up exactly the height of the screen This way, everything can be accessed easily throughout the application - no matter how large the content on the right side is (i.e. sticky/fixed menu on the left, content on the right).

Right now, one would have to scroll all the way down to access the footer element in the side menu/navbar.
image

Since the menu only has little content but still stretches to the bottom of the page, having the space on the left looks weird when editing a DMP. This would be fixed by having a sticky menu as well.
image

apps/damap-frontend/src/app/components/layout/layout.component.html
lpandath: Yes, that's the reason I asked on Discord, because having an icon was the quickest solution. In the first mockups, my solution was to only have the menu icon so that we could collapse everything at once or not (as it was in the old mobile version). However, after we decided to show the icons as Angular had in the documentation, as you mentioned, this now creates a different view. Therefore, I would suggest either leaving only the menu icon and making it expandable in the mobile version, or redirecting this icon, or perhaps adding more icons for different links.

lpandath: [DAMAP Dashboard.pdf](https://github.com/user-attachments/files/16813639/DAMAP.Dashboard.pdf)

apps/damap-frontend/src/app/components/layout/layout.component.html
lpandath: I have commented below!

apps/damap-frontend/src/app/components/layout/layout.component.html
rekt-hard: I was and still am in full support to simply be able to expand the menu on mobile as well. Then everything is accessible by expanding the menu and we do not have to come up with some extra logic only for mobile.

For the icon, maybe this one? I guess the admin panel icon will be used for the admin panel in the future.
image


We could add a tooltip with something like `Expand the menu to access legal documents and information`. What do you think? Also @DerekMolnar @ChristianeStork for text proposals

lpandath: p.s just to show the current changes. 



https://github.com/user-attachments/assets/00cee1da-83c6-4f27-bec8-e8037b520672

ChristianeStork: Thanks for the video. The suggested text is fine. I am just wondering if people will find the log-out button in the new position.

DerekMolnar: Yes, the video is helpful in seeing the changes.  In terms of verbiage, I suggest dropping "documents" so the tool tip would state: Expand the menu to access legal information.  I also think it would make more sense to have the logout icon higher up in the sidebar, e.g. below the DMPs icon.

lpandath: Final changes:


https://github.com/user-attachments/assets/5a93df6e-93e0-433c-9e39-0f336e00508e

https://github.com/user-attachments/assets/52a1d55a-76e4-4026-9ba4-171d48b9a70d

ValentinFutterer: Looks very good to me, just a few things from my side regarding the mobile version:
- The logout icon is not aligned with the other icons
- The info-icon in the footer should vanish when on mobile - otherwise people are going to try and expand the sidenav (which is not possible when on mobile)

rekt-hard: Maybe this got lost: Why should it not be possible to expand the menu on mobile? Users can close it, if they prefer to. If users are not able to expand the menu, they can not get to the legal information, which might be a legal issue in itself (not a legal expert though).

apps/damap-frontend/src/themes/custom-theme.scss
```diff
@@ -122,10 +123,27 @@ html {
 
   /* Support for IE. */
   font-feature-settings: "liga";
+
```
rekt-hard, line 126:
```suggestion
  // Set default icons to filled
  font-variation-settings: "FILL" 1;
  // class to set icons outlined. Example usage: <mat-icon class="outline">delete</mat-icon>
  &.outline {
    font-variation-settings: "FILL" 0;
  }
```

To make the default icon set filled instead of outlined. See also css on the right side: https://fonts.google.com/icons?selected=Material+Symbols+Outlined:delete:FILL@1;wght@400;GRAD@0;opsz@24&icon.size=24&icon.color=%23000000&icon.style=Outlined&icon.query=delete

libs/damap/src/lib/widgets/info-message/info-message.component.html
```diff
@@ -1,4 +1,4 @@
 <div class="info-message">
-  <mat-icon>info_outline</mat-icon>&nbsp;&nbsp;
+  <mat-icon>info</mat-icon>&nbsp;&nbsp;
```
rekt-hard:
```suggestion
  <mat-icon class="outline">info</mat-icon>&nbsp;&nbsp;
```

rekt-hard (commented):
Just a small icon issue. Before, we had an icon set which was filled per default.
Screenshot from 2024-09-02 16-23-08


With the new icon (or symbol) set, the default is outlined.
image


This can be changed though by adding the relevant css (see comment).

This means we should also update existing icons using the outline style (only used in a few places). From my understanding this means
`<mat-icon> icon_name_outline</mat-icon>` 

should be changed to
`<mat-icon class="outline">icon_name</mat-icon>`
image

apps/damap-frontend/src/app/components/layout/layout.component.ts
```diff
@@ -14,6 +20,7 @@ import pkg from '../../../../../../package.json'; // eslint-disable-line
 })
 export class LayoutComponent implements OnInit, AfterViewInit {
   @ViewChild('sidenav', { static: true }) sidenav!: MatSidenav;
+  @Input() info: string;
```
rekt-hard: Is this used?

apps/damap-frontend/src/app/components/layout/layout.component.html
```diff
@@ -70,23 +70,59 @@
...
+    <div *ngIf="isCollapsed" class="footer-icon">
       <button
         mat-icon-button
-        id="logout"
-        class="material-icon-on-secondary-container"
-        [attr.aria-label]="'layout.logout' | translate"
-        (click)="logout()">
-        <mat-icon>logout</mat-icon>
+        matTooltip="Expand the menu to access legal information"
+        aria-label=" Expand the menu to access legal information">
```
rekt-hard: would be great if these could be in the translation assets

apps/damap-frontend/src/themes/custom-theme.scss
lpandath: Yes, you can either add it like this, or as i removed in the tooltip the outlined part. Both works.

rekt-hard (approved):
looking good - thank you

PR #312, opened 2024-09-09
DMPStepper: Make evident if a section is locked

### LINKED ISSUE #35: make more evident if a section is locked, so that it might be recognised before selection
e.g. show an "x" icon instead of the default pencil.

it is important to still show the title of the step.

andresTabiTuwien:
Feature

## Description

Set the icon for when a step is blocked and requires action. 

https://github.com/tuwien-csd/damap-frontend/issues/35

#### What does this PR do?
Change the icons for steps in the stepper on dmpComponent where an action is required. For example, step 4 requires a dataset that was previously registered in step 3. 

Before
image

Now the icon when a step is blocked is a lock, and when a step is even empty, the step color remains gray. 

image

image

image

### Checks

- [ ] Summary updated
- [ ] Version view updated
- [ ] Documentation added
- [ ] Tests added
- [ ] E2e tests created

closes GH-35

### DIFF
--- a/apps/damap-frontend/src/styles.scss
+++ b/apps/damap-frontend/src/styles.scss
@@ -58,6 +58,9 @@
   // mat3 provides colors for error states, but not for success or warning...
   // so it is set to predefined values here
   --success-container: #e9f5d2;
+
+  --success-icon-color: #{mat.get-theme-color($app-light-theme, tertiary, 70)};
+
   --on-success-container: black;
   --warning-bg: #fff4cb;
   --background-body: #f6f6f6;
--- a/apps/damap-frontend/src/themes/custom-theme.scss
+++ b/apps/damap-frontend/src/themes/custom-theme.scss
@@ -157,6 +157,16 @@ html {
   @include mat.icon-color($app-light-theme, $color-variant: primary);
 }
 
+.material-icon-white-background-primary {
+  --mat-icon-color: white;
+  background-color: var(--primary);
+}
+
+.material-icon-white-background-success {
+  --mat-icon-color: white;
+  background-color: var(--success-icon-color);
+}
+
 .material-icon-on-secondary-container {
   --mat-icon-color: var(--on-secondary-container);
 }
--- a/libs/damap/src/lib/components/dmp/dmp.component.css
+++ b/libs/damap/src/lib/components/dmp/dmp.component.css
@@ -14,6 +14,21 @@
   margin-top: 10px;
 }
 
+.custom-icon-step {
+  border-radius: 50%;
+  padding: 4px;
+  width: 24px;
+  height: 24px;
+  display: flex;
+  justify-content: center;
+  align-items: center;
+}
+
+.number-step-icon {
+  font: var(--body-typography);
+  font-size: 1rem;
+}
+
 .verticalStepper {
   margin: 20px;
 }
--- a/libs/damap/src/lib/components/dmp/dmp.component.html
+++ b/libs/damap/src/lib/components/dmp/dmp.component.html
@@ -3,12 +3,108 @@
   #stepper
   [linear]="false"
   (selectionChange)="changeStep($event); changeStepPosition($event)">
-  <mat-step label="{{ 'dmp.steps.project.label' | translate }}">
+  <ng-template matStepperIcon="done" let-index="index">
+    <mat-icon
+      class="material-icon-white-background-success custom-icon-step"
+      *ngIf="iconsValidatorDone(index, 'check')"
+      >check</mat-icon
+    >
+    <mat-icon
+      class="material-icon-white-background-primary custom-icon-step"
+      *ngIf="iconsValidatorDone(index, 'edit')"
+      >edit</mat-icon
+    >
+    <mat-icon
+      class="custom-icon-step"
+      *ngIf="iconsValidatorDone(index, 'lock')"
+      >lock</mat-icon
+    >
+
+    <mat-icon
+      class="material-icon-white-background-success custom-icon-step"
+      *ngIf="iconsValidatorDone(index, 'text_snippet')">
+      text_snippet
+    </mat-icon>
+  </ng-template>
+
+  <ng-template matStepperIcon="edit" let-index="index">
+    <mat-icon
+      class="custom-icon-step"
+      *ngIf="iconsValidatorEdit(index, 'text_snippet')">
+      text_snippet
+    </mat-icon>
+
+    <mat-icon
+      class="material-icon-white-background-primary custom-icon-step"
+      *ngIf="iconsValidatorEdit(index, 'edit')">
+      edit
+    </mat-icon>
+
+    <mat-icon
+      class="material-icon-white-background-primary custom-icon-step"
+      *ngIf="iconsValidatorEdit(index, 'lock')"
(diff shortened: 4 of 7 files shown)


apps/damap-frontend/src/themes/custom-theme.scss
```diff
@@ -169,6 +174,11 @@ html {
   --mat-icon-color: darkorange;
 }
 
+.material-icon-white-background-warning {
+  --mat-icon-color: white;
+  background-color: darkorange;
```
ValentinFutterer: This would be a good ooportunity to create a new warning variable in `styles.csss`, to have the warning color fixed in one place.

libs/damap/src/lib/components/dmp/dmp.component.css
```diff
@@ -13,3 +13,13 @@
 .intro-step {
   margin-top: 10px;
 }
+
+.custom-icon-step {
+  border-radius: 50%; /* Make the background a circle */
+  padding: 8px; /* Add padding around the icon */
+  width: 24px; /* Adjust icon size */
+  height: 24px; /* Adjust icon size */
```
ValentinFutterer: Can you remove the ChatGPT comments :sweat_smile: :laughing:

ValentinFutterer:
Hey I have a few comments on this:
To my understanding, the goal was to mark a section as locked to prevent people from needlessly clicking on it. But with your implementation, the "locking" only triggers when a user clicks on a "locked" section, which defeats the whole purpose.

In the same vein, you should probably use another icon to signal "lockedness" (most likely a lock). Warning indicates that something is wrong or dangerous or should be looked at - but what we want to signal is, that you cant/shouldnt look at this part right now.

It would also be cool, if there is an icon for empty (suggestion just the grey number), one for filled out at least part of the step (suggestion pencil) and one for locked. Right now the pencil just stands for clicked on, which isnt really helping when navigating the tool. A clear seperation would help immensely.

Also, your icons are larger than the standard ones, which leads to a situation, where the icon that is in focus is smaller than its peers.
image

And another small thing - when you click on a "pristine" step, the icon changes to pencil, click on it again and it changes to number.

libs/damap/src/lib/components/dmp/dmp.component.ts
```diff
@@ -275,4 +283,53 @@ export class DmpComponent implements OnInit, OnDestroy {
...
+
+  get emptyReuse() {
+    return this.isFormGroupEmpty(this.reuseStep);
+  }
+
+  get emptyCost() {
+    return this.isFormGroupEmpty(this.costsStep);
+  }
+
+  isFormGroupEmpty(formGroup: UntypedFormGroup): boolean {
```
ValentinFutterer: Hey great work, i really love this new feature. I just noticed, that you could use the `summary.service.ts` service instead of checking the form group. This service is currently used in the summary step. This would have the added benefit of being able to differentiate between empty, worked on and done. Steps that are completely done could then be shown as green icons with a checkmark e.g. Sorry for not mentioning this sooner, I did not think about this at all when reviewing the last time 😅

andresTabiTuwien: Valen, I tested some green colors, but I'm not sure.
Here are the pictures of the tests.

color : --success-container
image

color: #94C11C <- This green is the same at the logo DMP in the app
image

Here I think there are many colors for the icons.

In this scenario, when a step is fully completed, the icon is the green check.
If a step is not fully complete but has some info, it is the blue pencil icon.
When a step has no info, the icon is the gray number.

image
Let me know what do you think about that.


Also I made the changes to use the sumary.service instead of check the form group.

ValentinFutterer: Using the tertiary color would also be an option - or you could create a green color palette - would prefer the first option though.

andresTabiTuwien: Yes, I updated the PR with the tertiary color.

apps/damap-frontend/src/themes/custom-theme.scss
```diff
@@ -169,6 +179,11 @@ html {
   --mat-icon-color: darkorange;
 }
 
+.material-icon-white-background-gray-color {
```
ValentinFutterer: This class is not needed, since these variables are the stepper default anyways.

ValentinFutterer (commented):
Summary step still has the default behaviour - i would suggest to replace the icon and to color it grey when nothing is filled out, blue pencil when something is filled out and turquoise when everything is done. 

When exporting a dmp and steps are not filled out, this warning message shows. Maybe you could reuse some logic here.

image

apps/damap-frontend/src/themes/custom-theme.scss
andresTabiTuwien: The class is needed because in order to personalize the icons in the stepper, the default behavior and variables of the step cannot remain as they are. 

provide: STEPPER_GLOBAL_OPTIONS,
useValue: { displayDefaultIndicatorType: false },
image

libs/damap/src/lib/components/dmp/dmp.component.html
```diff
@@ -3,12 +3,206 @@
   #stepper
   [linear]="false"
   (selectionChange)="changeStep($event); changeStepPosition($event)">
-  <mat-step label="{{ 'dmp.steps.project.label' | translate }}">
+  <ng-template matStepperIcon="done" let-index="index">
+    <mat-icon
```
lpandath, line 7: Works well! Would be great to consider refactoring the repeated *ngIf conditions by creating a helper function.

apps/damap-frontend/src/themes/custom-theme.scss
ValentinFutterer: Try deleting the class, you will observe that nothing changes. Look at the class closely, you are assigning the default stepper color values to the icon and background. Also now that i think about it, isnt the background gray and the icon white? Somethings not right here 😅
I think that `useValue: { displayDefaultIndicatorType: false },` just prevents the numerical icons from showing, the default colors still stay.

lpandath:
> Summary step still has the default behaviour - i would suggest to replace the icon and to color it grey when nothing is filled out, blue pencil when something is filled out and turquoise when everything is done.
> 
> When exporting a dmp and steps are not filled out, this warning message shows. Maybe you could reuse some logic here.
> image

@ValentinFutterer I did not touch the content of the steps as mention. So please create new issue for it. Thank you

apps/damap-frontend/src/styles.scss
```diff
@@ -58,6 +58,9 @@
   // mat3 provides colors for error states, but not for success or warning...
   // so it is set to predefined values here
   --success-container: #e9f5d2;
+
+  --success-icon-color: #{mat.get-theme-color($app-light-theme, tertiary, 70)};
```
ValentinFutterer, line 62: For me the color looks good, @lpandath do you want a different hue?

lpandath: _my opinion_:  As suggested in Discord (the differentiation between fill/outline) about the coloring. However, if you decide to choose an additional color, I would recommend using the lighter shades from the logo itself to ensure it fits well within the app.

PR #313, opened 2024-09-10
Gk/236 add admin only UI to work with internal storages

### LINKED ISSUE #236: Add admin only UI to work with internal storages
**Description**
Currently, internal storages can only be added via a sql script. This is annoying and a lot of extra work, if the list has to be changed.

The frontend should offer a admin-only UI element to perform CRUD operations on internal storages. This would make it way easier to manage them.

**Additional context**
This is closely related with https://github.com/tuwien-csd/damap-backend/issues/198 and https://github.com/tuwien-csd/damap-backend/issues/95 and should be worked on in tandem.

GeoffreyKarnbach:
## Description

Feature

#### What does this PR do?

Implementation of admin only UI to manage internal storages

#### Dependencies

Requires backend branch:
https://github.com/tuwien-csd/damap-backend/tree/gk/95-allow-internal-storages-to-be-disabled-for-the-frontend

### Checks

- [ ] Summary updated
- [ ] Version view updated
- [ ] Documentation added
- [ ] Tests added
- [ ] E2e tests created

closes GH-236

### DIFF
--- a/libs/damap/src/lib/components/admin/admin.component.ts
+++ b/libs/damap/src/lib/components/admin/admin.component.ts
@@ -1,0 +1,137 @@
+import { Component, OnInit } from '@angular/core';
+import { BackendService } from '../../services/backend.service';
+import { AuthService } from '../../auth/auth.service';
+import {
+  InternalStorage,
+  InternalStorageTranslation,
+} from '../../domain/internal-storage';
+import { MatDialog } from '@angular/material/dialog';
+import { InternalStorageDialogComponent } from './internal-storage-dialog/internal-storage-dialog.component';
+import { FeedbackService } from '../../services/feedback.service';
+import { Router } from '@angular/router';
+import { InternalStorageTranslationDialogComponent } from './internal-storage-translation-dialog/internal-storage-translation-dialog.component';
+import { firstValueFrom } from 'rxjs';
+
+@Component({
+  selector: 'damap-admin',
+  templateUrl: './admin.component.html',
+  styleUrl: './admin.component.css',
+})
+export class AdminComponent implements OnInit {
+  constructor(
+    private backendService: BackendService,
+    private dialog: MatDialog,
+    private feedbackService: FeedbackService,
+  ) {}
+
+  showOnlyActive = true;
+  internalStorages: InternalStorage[] = [];
+  internalStorageTranslations: InternalStorageTranslation[] = [];
+  selectedInternalStorageId: number;
+  selectedInternalStorageUrl: string;
+
+  ngOnInit(): void {
+    this.selectedInternalStorageId = null;
+    this.selectedInternalStorageUrl = null;
+
+    this.backendService.searchInternalStorage({}).subscribe(data => {
+      this.internalStorages = data.items;
+    });
+  }
+
+  async openStorageDialog() {
+    const dialogRef = this.dialog.open(InternalStorageDialogComponent, {
+      width: '75%',
+      maxWidth: '800px',
+      data: { mode: 'add' },
+    });
+
+    const storage = await firstValueFrom(dialogRef.afterClosed());
+    if (storage) {
+      this.backendService.createInternalStorage(storage).subscribe(
+        () => {
+          this.feedbackService.success('http.success.storage.add');
+          this.backendService.searchInternalStorage({}).subscribe(data => {
+            this.internalStorages = data.items;
+            this.selectStorage(data.items.find(s => s.url === storage.url).id);
+          });
+        },
+        error => {
+          if (error.error?.message) {
+            this.feedbackService.error(error.error.message);
+          } else {
+            this.feedbackService.error(error.message);
+          }
+        },
+      );
+    }
+  }
+
+  openTranslationDialog() {
+    const dialogRef = this.dialog.open(
+      InternalStorageTranslationDialogComponent,
+      {
+        width: '75%',
+        maxWidth: '800px',
+        data: { mode: 'add', storageId: this.selectedInternalStorageId },
+      },
+    );
+
+    dialogRef.afterClosed().subscribe(translation => {
+      if (translation) {
+        this.backendService
+          .createInternalStorageTranslation(translation)
+          .subscribe(
+            () => {
+              this.backendService
+                .getAllInternalStorageTranslationsForStorage(
+                  this.selectedInternalStorageId,
+                )
+                .subscribe(data => {
+                  this.internalStorageTranslations = data;
+                  this.feedbackService.success(
+                    'http.success.storage.translations.add',
+                  );
+                });
+            },
+            error => {
+              if (error.error?.message) {
+                this.feedbackService.error(error.error.message);
+              } else {
+                this.feedbackService.error(error.message);
+              }
+            },
+          );
+      }
+    });
+  }
+
+  selectStorage(storageId: number) {
+    if (storageId === null) {
+      this.resetStorageSelection();
+      return;
+    }
+    this.backendService.getInternalStorage(storageId).subscribe(storage => {
+      this.internalStorages = this.internalStorages.map(s =>
+        s.id === storageId ? storage : s,
+      );
+      this.backendService
+        .getAllInternalStorageTranslationsForStorage(storageId)
+        .subscribe(data => {
+          this.selectedInternalStorageId = storageId;
+          this.internalStorageTranslations = data;
+          this.selectedInternalStorageUrl = storage.url;
+        });
+    });
+  }
+
+  resetStorageSelection() {
+    this.selectedInternalStorageId = null;
+    this.internalStorageTranslations = [];
+    this.selectedInternalStorageUrl = null;
+
+    this.backendService.searchInternalStorage({}).subscribe(data => {
+      this.internalStorages = data.items;
+    });
+  }
+}
--- a/libs/damap/src/lib/components/dmp/data-storage/storage/storage.component.ts
+++ b/libs/damap/src/lib/components/dmp/data-storage/storage/storage.component.ts
@@ -48,8 +48,19 @@ export class StorageComponent implements OnInit {
   }
 
   private getInternalStorages() {
-    if (this.internalStoragesLoaded !== LoadingState.LOADED) {
-      this.store.dispatch(loadInternalStorages());
-    }
+    this.store.dispatch(loadInternalStorages());
+  }
+
+  get activeStorages() {
+    return this.internalStorages.filter(
+      storage => storage.active && storage.translations.length > 0,
+    );
+  }
+
+  public getStorageTitle(storage: InternalStorage) {
+    const translation = storage.translations.find(
+      t => t.languageCode === 'eng',
+    );
+    return translation ? translation.title : storage.translations[0].title;
   }
 }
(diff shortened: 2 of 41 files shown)


libs/damap/src/lib/components/admin/admin.component.ts
```diff
@@ -0,0 +1,132 @@
...
+  selectedInternalStorageId: number;
+  selectedInternalStorageUrl: string;
+
+  ngOnInit(): void {
+    this.selectedInternalStorageId = null;
+    this.selectedInternalStorageUrl = null;
+
+    if (!this.authService.isAdmin()) {
+      this.router.navigate(['/dashboard']);
+    }
```
rekt-hard: Might make sense to handle this via a route guard

libs/damap/src/lib/components/admin/admin.component.ts
```diff
@@ -0,0 +1,132 @@
...
+                data.items.find(s => s.url === storage.url).id,
+              );
+            });
+          },
+          error => {
+            this.feedbackService.error(error);
+          },
+        );
+      }
+    });
```
rekt-hard: I guess we could also await the first value from afterClosed, if we do not want to subscribe. We might have to manually unsubscribe. Not sure about that.

libs/damap/src/lib/components/dmp/data-storage/storage/storage.component.ts
```diff
@@ -48,8 +48,22 @@ export class StorageComponent implements OnInit {
   }
 
   private getInternalStorages() {
+    /*
     if (this.internalStoragesLoaded !== LoadingState.LOADED) {
       this.store.dispatch(loadInternalStorages());
     }
+      */
```
rekt-hard: Still needed?

libs/damap/src/lib/components/dmp/data-storage/storage/storage.component.ts
```diff
@@ -48,8 +48,22 @@ export class StorageComponent implements OnInit {
...
+    this.store.dispatch(loadInternalStorages());
+  }
+
+  get activeStorages() {
+    return this.internalStorages.filter(storage => storage.active);
+  }
+
+  public getStorageTitle(storage: InternalStorage) {
+    const translation = storage.translations.find(
+      t => t.languageCode === 'eng',
```
rekt-hard, line 62: It is really a shame that we have `eng` in the back end and `en` in the front end :/

libs/damap/src/lib/services/backend.service.ts
```diff
@@ -329,6 +334,89 @@ export class BackendService {
...
+    [key: string]: string[];
+  }): Observable<SearchResult<InternalStorage>> {
+    let params = new HttpParams();
+    for (const key in queryParams) {
+      if (queryParams.hasOwnProperty(key)) {
+        queryParams[key]?.forEach(item => (params = params.append(key, item)));
+      }
+    }
+    return this.http.get<SearchResult<InternalStorage>>(
+      `${this.backendUrl}storages/search`,
```
rekt-hard: should just be `/storages`

libs/damap/src/lib/services/form.service.ts
```diff
@@ -430,6 +438,76 @@ export class FormService {
...
+          Validators.maxLength(this.TEXT_SHORT_LENGTH),
+          notEmptyValidator(),
+        ],
+      ],
+      description: [
+        '',
+        [
+          Validators.required,
+          Validators.maxLength(this.TEXT_MAX_LENGTH),
+          notEmptyValidator(),
```
rekt-hard: Can be empty

libs/damap/src/lib/services/form.service.ts
```diff
@@ -430,6 +438,76 @@ export class FormService {
...
+        ],
+      ],
+      backupFrequency: [
+        '',
+        [
+          Validators.required,
+          Validators.maxLength(this.TEXT_SHORT_LENGTH),
+          notEmptyValidator(),
```
rekt-hard: can be empty

rekt-hard (commented):
Massive work - thank you!

Right now, we only have one page for the admin dashboard (as long as we do not shift the `All DMPs`component there). In general, it would be great to have a separate UI for the admin dashboard. On the left side, we could list the available sub pages (e.g. Internal Storages, DMPs, Users) and display the content on the right side. So at some point, this has to be refactored.

libs/damap/src/lib/services/backend.service.ts
```diff
@@ -388,32 +388,33 @@ export class BackendService {
     id: number,
   ): Observable<InternalStorageTranslation[]> {
     return this.http.get<InternalStorageTranslation[]>(
-      `${this.backendUrl}storages/${id}/translations`,
+      `${this.backendUrl}storages/${id}/translations/all`,
```
rekt-hard: major: To be consistent with other endpoints, I think we can omit the `/all` prefix for retrieving all entries or searching.
Also `/all` could be misunderstood as an identifier.

libs/damap/src/lib/services/backend.service.ts
```diff
@@ -388,32 +388,33 @@ export class BackendService {
...
   }
 
   getAllInternalStorageTranslationsForStorage(
     id: number,
   ): Observable<InternalStorageTranslation[]> {
     return this.http.get<InternalStorageTranslation[]>(
-      `${this.backendUrl}storages/translations/all/${id}`,
+      `${this.backendUrl}storages/${id}/translations/all`,
     );
   }
```
rekt-hard: major: can be removed as is the same as `getInternalStorageTranslations`

rekt-hard (commented):
Generally looking good. Just some small comments that should be quick to address.

One major issue though:
Getting an error when fetching all internal storages on admin page. Please also have a look at the back end.

libs/damap/src/lib/components/dmp/data-storage/storage/storage.component.ts
```diff
@@ -48,8 +48,17 @@ export class StorageComponent implements OnInit {
...
+    );
+    return translation ? translation.title : storage.translations[0].title;
```
rekt-hard, line 64: major: we should ensure that there is at least one translation before showing a storage option (so we might filter beforehand and never get to this point). Since it is a valid database state to have a storage without translations, this case should also be handled. From the back end standpoint, it will just return the storage with empty translations. The front end might will have to deal with this though. What do you think?

rekt-hard (commented):
1) major: In the admin panel, only the http status code is shown as an alert. Displaying the message sent with the response will help to better understand the issue (`Storage is still used in DMPs`).
Same for removing the last translation.
image

2) minor: Adapt the order and layout of the popup menu to reduce accidental clicks.
- List non-destructive actions first
- add separator/divider for destructive actions
- use color for destructive items to convey meaning. Either icons, buttons or text
image


3) major: When deleting a resource (storage, storage translation), showing a confirmation dialog could prevent accidental deletions.
For the dialog, having space between the actions reduces the possibility of clicking on the wrong action. Color will also help to convey meaning (red for delete, grey for cancel)
Example button layout:
`[ Delete (negative) | <space inbetween> | Cancel (standard/neutral) ]`

4) minor: When creating a new storage, the title of the storage translation is marked as required but it can be empty. When adding/updating a storage translation, the title is required to create/update.

libs/damap/src/lib/services/form.service.ts
```diff
@@ -351,9 +351,17 @@ export class FormService {
 
   public addStorageToForm(storage: InternalStorage) {
     const storageFormGroup = this.createStorageFormGroup();
+
+    const translation = storage.translations.find(
+      t => t.languageCode === 'eng',
+    );
+    const title: string = translation
+      ? translation.title
+      : storage.translations[0].title;
```
rekt-hard, line 364: see also general note on storage (there might not be a translation present)

libs/damap/src/lib/components/dmp/data-storage/storage/storage.component.ts
GeoffreyKarnbach: I actually do not agree on this point. Right now different validation mechanisms are in place to ensure that at least one translation should always exist (no have a title to be displayed in the frontend). Removing this condition would lead to additional (unnecessary) work, and users could for instance not remove a storage from their DMP anymore.

libs/damap/src/lib/widgets/internal-storage-table/internal-storage-table.component.ts
```diff
@@ -0,0 +1,144 @@
...
+                this.dataSource.data = this.internalStorages;
+                this.editTranslations(null);
+              },
+              error => {
+                if (error.status === 409) {
+                  this.feedbackService.error(
+                    this.translateService.instant(
+                      'http.error.storageErrors.stillInUse',
+                    ),
+                  );
```
rekt-hard, line 103: comment/shelve: This works for the current use case. In general, this will have to be adapted to show all messages from the back end. Might as well have them translated already in the back end, when user management is in place (so we can read the preferred language of the user) or when support for `accept-language` header is implemented.

rekt-hard (commented):
1) minor: when updating a translation and it fails, an http error is shown. Similar to other issues, a user-friendly message provides more context. This error was raised, when trying to add a translation with a language that already exists.
image

2) shelve: In general, we will want some sort of translation possibility in the back end. This will allow to supply the front end with already translated messages to be displayed. No additional error handling required for showing back end related messages in the front end. This will come in handy, when user management is in place (so we can read the preferred language of the user) or when support for accept-language header is implemented.

3) major: Can not update internal storage. The `update` button is always disabled. 
image

4) shelve: When a user updates a translations, showing some sort of success message would provide affirmative feedback. Feel free to just create an issue and postpone.

5) major: When updating a storage translation, a validation to check if there is already a translation for the selected language might be missing in the back end. This leads to the possibility of having two translations for the same language.
image

libs/damap/src/lib/widgets/internal-storage-translation-table/internal-storage-translation-table.component.ts
```diff
@@ -0,0 +1,153 @@
...
+                  t.id === translation.id ? translation : t,
+                );
+              this.dataSource.data = this.internalStorageTranslations;
+              this.feedbackService.success(
+                'http.success.storage.translations.edit',
+              );
+            },
+            error => {
+              if (error.error.message) {
+                this.feedbackService.error(error.error.message);
```
rekt-hard: can error be null/undefined and an error is raised for accessing `message` of it?
```suggestion
              if (error.error?.message) {
                this.feedbackService.error(error.error.message);
```

libs/damap/src/lib/widgets/internal-storage-translation-table/internal-storage-translation-table.component.ts
```diff
@@ -0,0 +1,153 @@
...
+              } else {
+                this.feedbackService.error(error.message);
+              }
```
rekt-hard: a short note on why this handling would help with providing some context

rekt-hard (commented):
1) major: Http error is displayed instead of a user centered error with details. This happens when trying to create a translation for a language that already exists. Probably just missing same error handling logic as in update method? 
image
Maybe makes sense to align these implementations at some point and reuse a common interface. 

2) major: Validator errors are not displayed correctly.
Screenshot from 2024-10-15 14-17-56
Expected display of message as seen in the manual contributor component: 
image

3) minor/shelve: When updating a translation, a success message is shown. For deleting a translation as well as for creating/updating/deleting a storage, nothing is shown. Similar visual feedback for users might be a good idea.

rekt-hard (approved):
UI looks good and is working great. Good job :clap:

PR #316, opened 2024-09-18
Sidenav: Change the location of a clase for the language btn in sidenav

### LINKED ISSUE #304: Sidenav: Use color system for button
**Changes**
`The class mm-navbar-button defines color for a mat-button, but this should be its own color class in custom-theme.scss. The html of the button that uses the mm-navbar-button class also has inline style - this should be removed.

**Acceptance Criteria**
No visible changes, but the code should be cleaner like described above.

andresTabiTuwien:
## Description

Change the place of a class that modify the color of the button for language in the sidenav.

[tuwien-csd/damap-frontend/issues/304](https://github.com/tuwien-csd/damap-frontend/issues/304)

### Checks

- [ ] Summary updated
- [ ] Version view updated
- [ ] Documentation added
- [ ] Tests added
- [ ] E2e tests created

closes GH-304

### DIFF
--- a/apps/damap-frontend/src/themes/custom-theme.scss
+++ b/apps/damap-frontend/src/themes/custom-theme.scss
@@ -169,6 +169,10 @@ html {
   --mat-icon-color: darkorange;
 }
 
+.material-btn-text-primary-lightest {
+  --mdc-text-button-label-text-color: var(--primary-lightest);
+}
+
 .snack-error {
   --mdc-snackbar-container-color: var(--error-container);
   --mdc-snackbar-supporting-text-color: var(--on-error-container);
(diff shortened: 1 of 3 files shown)


apps/damap-frontend/src/themes/custom-theme.scss
```diff
@@ -169,6 +169,10 @@ html {
   --mat-icon-color: darkorange;
 }
 
+.language-color-btn {
+  color: var(--primary-lightest) !important;
```
ValentinFutterer: This should be rewritten to `--mdc-text-button-label-text-color: var(--primary-lightest);`. Important is not necessary, you should always overwrite the internal color variables of mat3 components. 

You could also rename it. When we need a white button in the future, we can then just reuse this class.

ValentinFutterer (approved):
👍

PR #320, opened 2024-09-30
Live Preview: Show preview of export document in popup

### LINKED ISSUE #252: Preselection of the template, Real-time view
In many cases, the researchers already know which template they need. It would probably make sense to adapt the interface or query to these templates. Alternatively, a real-time view that exists in parallel to the query would be useful to see how it would look in the exported document (e.g. online latex editor Overleaf).

### LINKED ISSUE #259: Live DMP Export Preview
**Description**

When working on a DMP (either during development to try out template changes, or as a user), to check out the current export state, a Word Document has to be exported everytime.

In order to enhance productivity, a Live Preview of the DMP Export would be very useful.

**Proposed solution**

There are two approaches, that may both be explored for displaying the preview in the Frontend: 
- A split screen view (such as in overleaf) 
image
- An overlay approach (triggered through a button press to open a popup with the preview) 
image

Both approaches have advantages and disadvantages:
- The split screen view enables the user to always see the state of the preview (updated for instance when saving the DMP after making changes), but the usability on small screens might be impacted.
- The overlay allows better usability on smaller screens (especially for laptops) but the user has to press an additional button to be able to preview the current state.

In addition to the location in the frontend, different technical challenges regarding the rendering and backend exist:
- In the frontend, a DOCX to HTML renderer can be used to display (https://github.com/VolodymyrBaydalka/docxjs), no data has to be exposed to any third party services and the file can be kept private (behind an authentication wall), as it will only be used by an authenticated user within the frontend. Main issue with that approach is, that the rendering is not perfect and has some small flaws
image
- Alternatively, a live preview iframe vom Microsoft can be used. This renders the Document in a perfect manner, but leads to a public (see the next paragraph) available word document and it exposes it to Microsoft Services. Furthermore, this option is only available in a deployed environment, where the backend has a static URL (not localhost) that can be resolved my Microsoft services.
image

The approach to offer in localhost and in a deployed environment the live preview feature would be:
- Use the JS library to render the DOCX to HTML in the dev environment (not perfect, but probably good enough to check for quick updates made in the DMP Form)
- Use the Microsoft IFrame in deployment, ensuring a reliable rendering (and possible as the backend is hosted).

As mentioned before, the files need to be "publicly" available for the Microsoft Iframe to be able to access them. As the IFrame may also take API requests as input (as long as a Word Document is returned), following approach would lead to better security:
- Persist the generated preview Word Documents on the Server with a UUID as name (32 characters for instance)
- Add a OTP / Password (64 characters) that is passed as query param to the Backend call.
- Ensure that not used export files are deleted as soon as possible (keep track of the current filename in a DB entity, remove the file before generating a new export file for a given project).

Through this approach, the "public" files, are protected by a 32 characters UUID name (that would have to be guessed) and a 64 characters password within a "short" timeframe, making it effectively not feasible for an attacker (same would apply to any other endpoint if a JWT is leaked, the password has to be treated as an authentication/access token).

This approach would furthermore allow users to generate exported documents, that can be shared using an URL and the password as query param to any non authenticated user.

An approach to persisting those temporary files would be to the "temp" environment of the operating system (%TEMP% for windows, /var/tmp or /tmp for Unix).

**Alternative solution**
Generate a PDF on Server Side from the Word Document (no feasible approach was found, possible with libraries, but requires Microsoft Word to be installed in the Server Environment => not easily possible in Linux). This approach would make the entire approach easier, as rendering a PDF in the Frontend is trivial.

GeoffreyKarnbach:
## Description

Display a rendered version of the current DMP in DOCX, displayed as PDF using an embed.

#### What does this PR do?

Add new button in the action bar to display the preview popup.

#### Dependencies

https://github.com/tuwien-csd/damap-backend/pull/260

### Checks

- [ ] Summary updated
- [ ] Version view updated
- [ ] Documentation added
- [ ] Tests added
- [ ] E2e tests created

### DIFF
--- a/libs/damap/src/lib/components/dmp/live-preview/live-preview.component.html
+++ b/libs/damap/src/lib/components/dmp/live-preview/live-preview.component.html
@@ -1,0 +1,45 @@
+<div class="dialog-header">
+  <h1 mat-dialog-title>{{ "preview.title" | translate }}</h1>
+  <button mat-button mat-dialog-close class="top-right-button">
+    <mat-icon style="margin: 0">close</mat-icon>
+  </button>
+</div>
+
+<mat-dialog-content class="full-height max-height-90">
+  <div class="container full-height">
+    <p>
+      {{ "preview.info" | translate }}
+    </p>
+    <ng-container>
+      <div class="select-row">
+        <mat-form-field class="expanded-select" appearance="fill">
+          <mat-label>{{ "preview.templateSelection" | translate }}</mat-label>
+          <mat-select
+            [(value)]="selectedTemplate"
+            (selectionChange)="onTemplateChange($event)">
+            <mat-option
+              *ngFor="let id of dmpTemplate | keyvalue"
+              [value]="id.key">
+              {{ id.value | translate }}
+            </mat-option>
+          </mat-select>
+        </mat-form-field>
+      </div>
+    </ng-container>
+    <br />
+    <div *ngIf="pdfUrl" style="height: 80vh">
+      <iframe
+        [src]="pdfUrl"
+        width="100%"
+        height="100%"
+        title="preview"></iframe>
+    </div>
+    <div *ngIf="!pdfUrl && selectedTemplate.length !== 0" class="no-preview">
+      <mat-icon class="loading-icon">autorenew</mat-icon>
+      <p>{{ "preview.info" | translate }}</p>
+    </div>
+    <div *ngIf="!pdfUrl && selectedTemplate.length === 0" class="no-preview">
+      <p>{{ "preview.requestTemplateSelection" | translate }}</p>
+    </div>
+  </div>
+</mat-dialog-content>
--- a/libs/damap/src/assets/i18n/templates/en.json
+++ b/libs/damap/src/assets/i18n/templates/en.json
@@ -468,7 +468,11 @@
     "save": "Save changes",
     "discard": "Discard changes",
     "version": "Save new version",
-    "export": "Export document"
+    "export": "Export document",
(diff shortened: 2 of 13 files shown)


libs/damap/src/lib/components/dmp/live-preview/live-preview.component.html
```diff
@@ -0,0 +1,45 @@
+<div class="dialog-header">
+  <h1 mat-dialog-title>{{ "preview.title" | translate }}</h1>
+  <button mat-button mat-dialog-close class="top-right-button">
+    <mat-icon style="margin: 0">close</mat-icon>
+  </button>
+</div>
+
+<mat-dialog-content style="height: 100%; max-height: 90%">
+  <div class="container" style="height: 100%">
```
rekt-hard: shall we move all inline styling to the respective css files and just use an appropriate class here?

rekt-hard (commented):
Awesome feature - thank you :rocket: 

Would it be feasible to show the preview and still interact with the site? Then update the preview when the `update dmp` request succeeds. So the live preview component would then have to be draggable and not a complete overlay like a dialog, so users can still interact with the site without closing the previewer.

GeoffreyKarnbach:
I made a new issue for an option to make the preview window draggable, as discussed.

https://github.com/tuwien-csd/damap-frontend/issues/348

rekt-hard (approved):
Works great - good job!
Can be rebased and then merged, after the back end PR has been merged.

rekt-hard (approved):
Good job! LGTM

PR #334, opened 2024-10-09
global: Use default behaviour for changedetector

### LINKED ISSUE #333: Table does not automatically update
**Describe the bug**
When uploading a file to be analyzed, the UI does not get updated automatically. An interaction with another UI element is necessary to trigger the re rendering.

The issue is present in the commit for v4.2.0 all the way to commit [f3772752c269ba76193e3d8e99a86c9f5649cc5a](https://github.com/tuwien-csd/damap-frontend/commit/f3772752c269ba76193e3d8e99a86c9f5649cc5a). So the assumption is that it was introduced in this commit.

With this commit, also the highlighting of the selected tab was removed.

**Is this a regression?**
Yes

**To Reproduce**

1. Create a DMP
2. Go to section 3
3. Click on `Specify dataset` for new data
4. Click on `Upload sample data`
5. Select a file for upload
6. Observe that the progress bar does not update and the result will not be added to the table.

**Expected behavior**
Progress bar gets updated.
Result of uploaded file is added to the table.

**Screenshots**
***Before***
- Screenshot from 2024-10-09 18-45-44
- Screenshot from 2024-10-09 18-46-21
- Screenshot from 2024-10-09 18-46-32

***Now***
- Screenshot from 2024-10-09 18-47-40
- After interacting with the file upload again
- image

**Environment**

- Browser: firefox
- Damap version: v4.2.0 to commit f3772752c269ba76193e3d8e99a86c9f5649cc5a

**Additional context**

rekt-hard:
#### What does this PR do?
Use default change detection for components.
For tables, no change was detected when updating data in the background. It would have required a manual push to render again.

### Checks

- [ ] Summary updated
- [ ] Version view updated
- [ ] Documentation added
- [ ] Tests added
- [ ] E2e tests created

closes #333

### DIFF
--- a/apps/damap-frontend/src/app/components/layout/layout.component.ts
+++ b/apps/damap-frontend/src/app/components/layout/layout.component.ts
@@ -1,6 +1,5 @@
 import {
   AfterViewInit,
-  ChangeDetectionStrategy,
   Component,
   OnInit,
   ViewChild,
@@ -19,7 +18,6 @@ import { BehaviorSubject, filter, Subscription } from 'rxjs';
 import * as layoutTemplate from '../../../assets/i18n/layout/en.json';
 
 @Component({
-  changeDetection: ChangeDetectionStrategy.OnPush,
   selector: 'app-layout',
   templateUrl: './layout.component.html',
   styleUrls: ['./layout.component.css'],
--- a/apps/damap-frontend/src/app/components/layout/layout.component.ts
+++ b/apps/damap-frontend/src/app/components/layout/layout.component.ts
@@ -7,14 +7,14 @@ import {
 } from '@angular/core';
 import { BreakpointObserver, Breakpoints } from '@angular/cdk/layout';
 
-import { AuthService } from '@damap/core';
+import { AuthService, DashboardComponent } from '@damap/core';
 import { ConfigService } from '../../services/config.service';
 import { MatSidenav } from '@angular/material/sidenav';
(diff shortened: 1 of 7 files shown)


apps/damap-frontend/src/app/components/layout/layout.component.ts
```diff
@@ -37,8 +35,7 @@ export class LayoutComponent implements OnInit, AfterViewInit, OnDestroy {
   public isCollapsed: boolean = false;
   public templateDataLayout = layoutTemplate;
   public icon = 'info';
-  public infoString: BehaviorSubject<any> | null;
```
rekt-hard, line 40: Removed as it was not used (only initialized)

apps/damap-frontend/src/app/components/layout/layout.component.ts
```diff
@@ -64,8 +61,9 @@ export class LayoutComponent implements OnInit, AfterViewInit, OnDestroy {
     this.observer.observe([Breakpoints.Handset]).subscribe(result => {
       this.isSmallScreen = result.matches;
       this.checkScreenSize();
-      this.handleRouteChange();
```
rekt-hard, line 67: This should only be relevant when the route changes, which we are already subscribed to

andresTabiTuwien (approved):
Thank you, the refactor for the handleRouteChange function and the removal of the onPush strategy fixes the problem.

PR #335, opened 2024-10-11
dependencies: Update to nx 19.8.4 and angular 18.2.8

rekt-hard:
general notes:
* remove experimental dependency
* remove explicit zone.js (included in angular)
* run nx migrations

commands run:
- npx nx migrate 19
- npm install
- npx nx migrate --run-migrations


ValentinFutterer:
This update seems to work for Linux, but it fails for Windows. So a Windows user needs to change some dependencies that are causing issues. (Waiting could also just solve it, since newer versions could have fixed the bug)

ValentinFutterer:
I updated again, but the bug is still not fixed - now it works on Windows but fails in the pipeline :/
Others are experiencing similar [problems](https://github.com/nrwl/nx/issues/28205).
We should probably wait even longer and hope for a fix.

PR #337, opened 2024-10-15
🔗 licenses: make URLs consistent with damap-backend

SotosTsepe:


ValentinFutterer (approved):
Thanks for the help, url's match with tthe ones in the frontend.

PR #342, opened 2024-10-17
Username: Fix the undefined when the user name is not set

### LINKED ISSUE #341: User name is show as a undefined
If the user does not have a first name in keycloak, the username will show up as undefined in DAMAP.
Image
Should not show undefined if the user does not have the firstname

andresTabiTuwien:
### Checks

- [ ] Summary updated
- [ ] Version view updated
- [ ] Documentation added
- [ ] Tests added
- [ ] E2e tests created

closes GH-

### DIFF
--- a/apps/damap-frontend/src/app/components/layout/layout.component.ts
+++ b/apps/damap-frontend/src/app/components/layout/layout.component.ts
@@ -54,7 +54,7 @@ export class LayoutComponent implements OnInit, AfterViewInit, OnDestroy {
   }
 
   ngOnInit(): void {
-    this.name = this.auth.getName();
+    this.name = this.auth.getName() ?? '';
     const browserLang = this.translate.getBrowserLang();
     this.translate.use(browserLang?.match(/en|de/) ? browserLang : 'en');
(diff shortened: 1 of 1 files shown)


apps/damap-frontend/src/app/components/layout/layout.component.ts
```diff
@@ -54,7 +54,7 @@ export class LayoutComponent implements OnInit, AfterViewInit, OnDestroy {
   }
 
   ngOnInit(): void {
-    this.name = this.auth.getName();
+    this.name = this.auth.getName() ? this.auth.getName() : '';
```
rekt-hard: Using the null coalescing operator allows to call the method only once. If the left operand returns a value other than `null` or `undefined`, this value will be assigned - otherwise the right operand will be assigned.
```suggestion
    this.name = this.auth.getName() ?? '';
```

andresTabiTuwien: Thanks David, I corrected it.

rekt-hard (approved):
good catch - thank you

PR #359, opened 2024-11-06
Internal Storage: Add URL validation using "validator" package

### LINKED ISSUE #353: Internal Storage: URL is not checked for well-formedness
**Describe the bug**
You can add `internal gitlab` as URL and save the internal storage. This causes an error while exporting to madmp json, since it expects a well formed URL. Add a check in the frontend to prevent this from happening. Related backend issue: https://github.com/tuwien-csd/damap-backend/issues/276

Also check against the [RDA-common-standard](https://github.com/RDA-DMP-Common/RDA-DMP-Common-Standard?tab=readme-ov-file#host_storage_type_tree) if anything else needs to be validated.

**Is this a regression?**
No.

**Expected behavior**
Adding a malformed URL should show an error state.

**Environment**

- Damap version 4.2.2

GeoffreyKarnbach:
## Description

Add URL frontend validation using the "validator" package.

closes GH-353

### DIFF
--- a/libs/damap/src/lib/components/admin/admin.component.ts
+++ b/libs/damap/src/lib/components/admin/admin.component.ts
@@ -11,6 +11,7 @@ import { FeedbackService } from '../../services/feedback.service';
 import { Router } from '@angular/router';
 import { InternalStorageTranslationDialogComponent } from './internal-storage-translation-dialog/internal-storage-translation-dialog.component';
 import { firstValueFrom } from 'rxjs';
+import validator from 'validator';
 
 @Component({
   selector: 'damap-admin',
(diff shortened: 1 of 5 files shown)


libs/damap/src/lib/components/admin/admin.component.ts
```diff
@@ -48,6 +49,10 @@ export class AdminComponent implements OnInit {
 
     const storage = await firstValueFrom(dialogRef.afterClosed());
     if (storage) {
+      if (!this.isValidUrl(storage.url)) {
+        this.feedbackService.error('http.error.storage.invalidUrl');
```
ValentinFutterer: Is it possible that you forgot to add this translation to the translation files 😜

GeoffreyKarnbach: Nice catch :)

ValentinFutterer (approved):
Well done!

PR #362, opened 2024-11-11
Fix the update project and update acronym labels for project name and…

### LINKED ISSUE #325: Step 1 Changes: Remove Project Description, add Funder and Grant number
**Changes**
1. In Step 1 remove the “Project description” field under “Input project manually”.  Rationale: the content from the field does not export to the template and is not required by the funders.  
2. In place of the field, add a free text field for “Funder” and a free text field for “Grant number” Neither should be required.  This information should go into the “Funder, funding programme, grant number” section of the Project Details table in the Science Europe template. Under the FWF grant information in the first table.  For the HE template, it should go in the field for the grant number in the footer of the cover page and in the Copyright Notice (see screenshot) 
**Consider adding a dropdown menu for the funder (FWF, EU – Horizon Europe, Other) to automatically choose the correct template**
3. The name of “Update project” should be changed to “Project name”;  “Update acronym” should be changed to “Project acronym”.
**Screenshot**
Image
Image
Image

andresTabiTuwien:
… project acronym add fields for funder and grant

## Description

Add two fields for manual entry of the project, the funder and grant fields.  For each of these there is also a drop down box to select the type of data for the fields.

image

https://github.com/tuwien-csd/damap-frontend/issues/325

### Checks

- [ ] Summary updated
- [ ] Version view updated
- [ ] Documentation added
- [ ] Tests added
- [ ] E2e tests created

closes GH-325

### DIFF
--- a/libs/damap/src/assets/i18n/templates/en.json
+++ b/libs/damap/src/assets/i18n/templates/en.json
@@ -6,9 +6,13 @@
           "title": "Project title",
           "description": "Project description",
           "button": "Update project",
-          "acronym": "Update acronym",
+          "acronym": "Project acronym",
+          "funderID": "Funder",
+          "funderType": "Funder ID Type",
           "date": "Enter start and end date",
-          "datehint": "DD/MM/YYYY – DD/MM/YYYY"
+          "datehint": "DD/MM/YYYY – DD/MM/YYYY",
+          "grantID": "Grant ID",
+          "grantType": "Grant ID Type"
         },
         "manual.input": "Input project manually",
         "list": "Choose project from the project database",
--- a/libs/damap/src/lib/components/dmp/project/manual-project-input/manual-project-input.component.html
+++ b/libs/damap/src/lib/components/dmp/project/manual-project-input/manual-project-input.component.html
@@ -3,7 +3,7 @@
 }}</mat-label>
 <mat-card class="manual-project-input-card card-shadow">
   <mat-card-content>
-    <form>
+    <form [formGroup]="form">
       <div class="form-row">
         <app-input-wrapper
           class="input-full"
@@ -42,6 +42,55 @@
         label="{{
           'dmp.steps.project.input.description' | translate
         }}"></app-textarea-wrapper>
+
+      <div formGroupName="funding">
+        <div class="form-row" formGroupName="funderId">
+          <app-input-wrapper-options
+            class="input-60"
+            [control]="funder"
+            [options]="funderOptions"
+            (inputChange)="selectOptionFunderId($event)"
+            label="{{
+              'dmp.steps.project.input.funderID' | translate
+            }}"></app-input-wrapper-options>
+          <mat-form-field>
+            <mat-label>{{
+              "dmp.steps.project.input.funderType" | translate
+            }}</mat-label>
+            <mat-select formControlName="type">
+              <mat-option
+                *ngFor="let type of funderTypeObject | keyvalue"
+                [value]="type.value">
+                {{ type.value | translate }}
+              </mat-option>
+            </mat-select>
+          </mat-form-field>
+        </div>
+
+        <div class="form-row" formGroupName="grantId">
+          <app-input-wrapper
+            class="input-60"
+            [control]="grant"
+            (inputChange)="checkValidators()"
+            label="{{
+              'dmp.steps.project.input.grantID' | translate
+            }}"></app-input-wrapper>
+
+          <mat-form-field>
+            <mat-label>{{
+              "dmp.steps.project.input.grantType" | translate
+            }}</mat-label>
+            <mat-select formControlName="type">
+              <mat-option
+                *ngFor="let type of grantTypeObject | keyvalue"
+                [value]="type.value">
+                {{ type.value | translate }}
+              </mat-option>
+            </mat-select>
+          </mat-form-field>
+        </div>
+      </div>
+
       <button
         class="button-position-left button-color-primary"
         mat-raised-button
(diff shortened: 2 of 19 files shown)


libs/damap/src/lib/services/template-selector.service.ts
```diff
@@ -0,0 +1,45 @@
...
+import { ETemplateType } from '../domain/enum/export-template-type.enum';
+import { EFunderIdsHelper } from '../domain/enum/fundersIds.enum';
+
+@Injectable({
+  providedIn: 'root',
+})
+export class TemplateSelectorService {
+  constructor() {}
+
+  selecTemplate(dmp: Dmp): ETemplateType {
```
ValentinFutterer: Small typo

libs/damap/src/lib/components/dmp/project/manual-project-input/manual-project-input.component.html
```diff
@@ -28,6 +28,50 @@
...
+
+  <div formGroupName="funding">
+    <div class="form-row" formGroupName="funderId">
+      <app-input-wrapper-options
+        class="input-60"
+        [control]="funder"
+        [options]="funderOptions"
+        label="{{ 'Funder' | translate }}"></app-input-wrapper-options>
+
+      <mat-form-field appearance="fill">
```
ValentinFutterer: Non-textarea form fields should have appearance="outline" to keep in sync with the rest of the project.

libs/damap/src/lib/components/dmp/project/manual-project-input/manual-project-input.component.html
```diff
@@ -28,6 +28,50 @@
...
+      </mat-form-field>
+    </div>
+
+    <div class="form-row" formGroupName="grantId">
+      <app-input-wrapper
+        class="input-60"
+        [control]="grant"
+        label="{{ 'Grant number' | translate }}"></app-input-wrapper>
+
+      <mat-form-field appearance="fill">
```
ValentinFutterer: Non-textarea form fields should have appearance="outline" to keep in sync with the rest of the project.

libs/damap/src/lib/widgets/export-warning-dialog/export-warning-dialog.html
```diff
@@ -3,7 +3,7 @@ <h1 mat-dialog-title>{{'dialog.export.title' | translate}}</h1>
   {{'dialog.export.content' | translate}}
   <div class="container">
     <ng-container>
-      <mat-form-field *ngIf="!funderSupported">
+      <mat-form-field appearance="fill">
```
ValentinFutterer: Non-textarea form fields should have appearance="outline" to keep in sync with the rest of the project.

libs/damap/src/assets/i18n/templates/en.json
```diff
@@ -5,10 +5,12 @@
         "input": {
           "title": "Project title",
           "description": "Project description",
-          "button": "Update project",
-          "acronym": "Update acronym",
+          "button": "Project name",
```
ValentinFutterer: This key affects the button and the label of the name input field - only the label should be chnaged to project name though

libs/damap/src/lib/components/dmp/project/manual-project-input/manual-project-input.component.html
```diff
@@ -28,6 +28,50 @@
...
       'dmp.steps.project.input.description' | translate
     }}"></app-textarea-wrapper>
+
+  <div formGroupName="funding">
+    <div class="form-row" formGroupName="funderId">
+      <app-input-wrapper-options
+        class="input-60"
+        [control]="funder"
+        [options]="funderOptions"
+        label="{{ 'Funder' | translate }}"></app-input-wrapper-options>
```
ValentinFutterer: There is no translation key "Funder". It should say "Funder ID", to make it more obvious for users what to put in this field.

ValentinFutterer:
Very good, i would just add some minor functionality changes

Selecting a funder id over dropdown should also preselect the type - otherwise a user has no idea what kind of ID this is.
image

Also, if a funder id is selected, it should be mandatory to select an id type and vice-versa, same for grant_is. See the [RDA-Common standard](https://github.com/RDA-DMP-Common/RDA-DMP-Common-Standard?tab=readme-ov-file#properties-in-grant_id)

Small info: The previous implementation was that the user should have a template preselected, only if there is funding. Otherwise nothing should be preselected.
image

For some weird reason, these values are lowercase, the enum is in upper case though.
image

libs/damap/src/assets/i18n/templates/en.json
```diff
@@ -5,10 +5,12 @@
...
+          "acronym": "Project acronym",
+          "funderType": "Funder Type",
           "date": "Enter start and end date",
-          "datehint": "DD/MM/YYYY – DD/MM/YYYY"
+          "datehint": "DD/MM/YYYY – DD/MM/YYYY",
+          "granType": "Grant Type"
```
ValentinFutterer: Small typo

libs/damap/src/lib/components/dmp/project/manual-project-input/manual-project-input.component.html
```diff
@@ -28,6 +28,50 @@
...
+          </mat-option>
+        </mat-select>
+      </mat-form-field>
+    </div>
+
+    <div class="form-row" formGroupName="grantId">
+      <app-input-wrapper
+        class="input-60"
+        [control]="grant"
+        label="{{ 'Grant number' | translate }}"></app-input-wrapper>
```
ValentinFutterer: Same as above, no translation key.

ValentinFutterer:
Almost everything works now, but there are still a few small things missing:

- One can choose an ID type, without being forced to also input the actual ID.
image

- Inputing just the id's blocks the button, like intended. But there is no immediate user feedback, except for the star next to id type.
image

- There are some verbiage issues. Instead you should use `"New project"`, `"Project title"` and `"Funder ID"`.
image

- We should check if the ids follow their specific id structure determined by the chosen id type. If an id doesnt follow the exact required structure, the user should be made aware of this. There is already a URI check in the frontend, you can reuse this for URL. For the others you will probably need to write regex. These can then be later reused in the backend.

ValentinFutterer:
Just checked the RDA recommendation - a project can have [multiple fundings](https://github.com/RDA-DMP-Common/RDA-DMP-Common-Standard?tab=readme-ov-file#properties-in-project). We need to cover this case with the implementation.

ValentinFutterer:
Christiane: Says description input should be removed from manual input or reduced to 200 words (at least for now, when we still export to Word). She would like to have the option to input the actual name of funders (dropdown + freetext) and not just ids. Inputing just IDs is not very userfriendly and we should stray away.

Adding fundername would require us to extend the RDA - CS

PR #366, opened 2024-11-13
UI: Add undefined check for version info for older backend versions in the DMP Table

### LINKED ISSUE #360: Plans: Handle null/undefined version tag
**Describe the bug**
When the version is null or undefined, this happens:
image

**Is this a regression?**
No.

**Expected behavior**
It should be handled by leaving it blank, or with a text explaining that there is no version currently.

**Environment**

- Damap version 4.2.2

GeoffreyKarnbach:
## Description

Add check for undefined in the DMP Table Version Column (relevant if older backend version is used). Add verbiage to redirect to version page if no version is provided by the backend.

closes GH-360

### DIFF
--- a/libs/damap/src/lib/widgets/dmp-table/dmp-table.component.html
+++ b/libs/damap/src/lib/widgets/dmp-table/dmp-table.component.html
@@ -39,21 +39,29 @@
       <td mat-cell *matCellDef="let element">
         <a class="version" routerLink="/dmp/{{ element.id }}/version">
           {{
-            element.versionCount !== null
+            element.versionCount !== null && element.versionCount !== undefined
               ? ("plans.table.header.version_tag" | translate) + ": "
               : ""
           }}
           <span class="underlined">{{
-            element.versionCount !== null ? "#" + element.versionCount : ""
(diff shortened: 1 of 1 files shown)


libs/damap/src/lib/widgets/dmp-table/dmp-table.component.html
```diff
@@ -39,21 +39,25 @@ <h3 matTooltip="Edit DMP">
...
+              ? "#" + element.versionCount
+              : ""
           }}</span>
           <br />
           {{
-            element.versionCount !== null
+            element.versionCount !== null && element.versionCount !== undefined
               ? ("plans.table.header.version_name" | translate) + ": "
-              : ""
+              : "Check out the version page"
```
ValentinFutterer: Why "Check out the version page"? This text is displayed when no version is available. When you click on it, it brings you to an empty version table. It should probably say "No versions for this DMP available".

And now that i look at it more closely - the `<br />` is always displayed, even there are no versions. It should be pulled into the conditional.

ValentinFutterer (approved):
Nice

PR #376, opened 2024-11-22
Improve: add upload dialog with DatasetDialogUploadComponent for file uploads

### LINKED ISSUE #375: Datastep: Mat-Tab-Group only shows one header
**Describe the bug**
This happens when switching between new data and reused data.

Should be:
image

But is:
image

I returns to normal when you click on it.

**Is this a regression?**
Yes.

**Expected behavior**
Should look like the first screenshot from the beginning.

**Environment**

- Browser Chrome
- Damap version 4.3.0

**Discussion**
`Add dataset manually` opens a popup and `Upload sample data` is embedded in the normal flow of the app. This has bothered me for some time now.
Maybe we should remove the tab-group here altogether. As an idea: `Add dataset manually` and `Upload sample data` could just be two buttons next to each other. `Add dataset manually` already opens a popup, so why not `Upload sample data` too?

Maybe you have different ideas?

lpandath:
## Description
Screenshot 2024-12-03 at 13 56 59

#### What does this PR do?

This PR introduces a new `DatasetDialogUploadComponent` to handle file uploads in a dedicated dialog. 

#### Code review focus

- Ensure the `DatasetDialogUploadComponent` is properly integrated and functions as expected.
- Verify dialog closure logic in `DatasetDialogUploadComponent`.

### Checks

- [ ] Summary updated
- [ ] Version view updated
- [ ] Documentation added
- [ ] Tests added
- [ ] E2e tests created

closes GH-

### DIFF
--- a/apps/damap-frontend/src/app/services/config.service.ts
+++ b/apps/damap-frontend/src/app/services/config.service.ts
@@ -1,17 +1,18 @@
 import { AuthConfig, OAuthService } from 'angular-oauth2-oidc';
+import { BehaviorSubject, Observable, lastValueFrom } from 'rxjs';
 import { Injectable, isDevMode } from '@angular/core';
 
 import { Config } from '@damap/core';
 import { HttpClient } from '@angular/common/http';
-import { lastValueFrom } from 'rxjs';
-import { environment } from '../../environments/environment';
 import { Router } from '@angular/router';
+import { environment } from '../../environments/environment';
 
 @Injectable({
   providedIn: 'root',
 })
 export class ConfigService {
   private config: Config;
+  private configSubject = new BehaviorSubject<Config | null>(null);
 
   constructor(
     private http: HttpClient,
@@ -28,6 +29,7 @@ export class ConfigService {
           return new Promise<boolean>(resolve => resolve(false));
         } else {
           this.config = config;
+          this.configSubject.next(config);
           const authConfig: AuthConfig = {
             issuer: config.authUrl,
             clientId: config.authClient,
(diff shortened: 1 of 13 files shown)


ValentinFutterer:
Functionality is all there, the new upload dialog makes the application more consistent and user friendly. 

Before we can merge it, I would like to address two things:

- The bug mentioned in the issue still occurs. This is probably caused by the fitsServiceAvailable variable not being loaded correctly.

image

- If we want to change our main button type to FAB, we have to keep it consistent in the whole application. This should be discussed with the whole team. I am in favor of not changing our standard button design. The main purpose of FAB buttons is to enable the main action of an app, like save e.g. Read more here in the [styleguide](https://m3.material.io/components/floating-action-button/overview). 
image

lpandath:
> Functionality is all there, the new upload dialog makes the application more consistent and user friendly.
> 
> Before we can merge it, I would like to address two things:
> 
>     * The bug mentioned in the issue still occurs. This is probably caused by the fitsServiceAvailable variable not being loaded correctly.
> 
> 
> image
> 
>     * If we want to change our main button type to FAB, we have to keep it consistent in the whole application. This should be discussed with the whole team. I am in favor of not changing our standard button design. The main purpose of FAB buttons is to enable the main action of an app, like save e.g. Read more here in the [styleguide](https://m3.material.io/components/floating-action-button/overview).
> 
> 
> image

@ValentinFutterer,
Thank you for your feedback! Here’s my input regarding the points mentioned:

- Bug with fitsServiceAvailable:
I’ll investigate this further to ensure that the fitsServiceAvailable variable is loaded correctly and the issue does not persist. Maybe we can have a meeting for it. 
- FAB Button Design:
 already posted in our chat group about this decision, and your reaction indicated agreement. But happy to revisit this topic again.

ValentinFutterer:
About the preview bug, maybe you can take inspiration from this: https://github.com/tuwien-csd/damap-frontend/pull/392
Yes i didnt look at it in depth back then. I researched FAB buttons today, i didnt know about this either.

andresTabiTuwien (commented):
Good job, the bug is fixed and the buttons are consistent with the app.

andresTabiTuwien (approved):

PR #394, opened 2024-12-02
List contributors: Display the contributors in both steps and add an …

### LINKED ISSUE #390: Contributors: Rework UI
**Changes**
The second button does not make any sense in the context of the new UI - it should say `Input contributor manually`. Similarily, the `Hide manual contributor input` can be removed, since it does not serve any purpose in the new UI. 
image

Additionally, I would suggest to use a similar layout to step 8 and show the chosen contributors in both steps, so the user has a good overview, regardless to which tab he/she has navigated. 

A common issue that many users have faced, is that the contact person button/icon is very, very hidden in the UI. mayn simply miss it, or misuse it. It also causes misalignment in the  contributor cards.

image

To fix this, I would suggest to create a specific contact person "slot" in the list, that is visually distinct from the other ones. The `Set as contributor` and `Set as contact person button` should get some text next to thew icons, like `Contributor` and `Main Contact`, to make it clear what they down on first glance.
image

image

Other solution ideas are also encouraged, be creative about it!

**Acceptance Criteria**
The tab button should be renamed and the hide button should be removed. Additonally, the contact person system should be reworked in a way as to allow users more easily to grasp the concept.

andresTabiTuwien:
…info box to guide the user to select a contact person.

## Description

List contributors: Display the contributors in both steps and add an info box to guide the user to select a contact person.

https://github.com/tuwien-csd/damap-frontend/issues/390
#### What does this PR do?
image
image

### Checks

- [ ] Summary updated
- [ ] Version view updated
- [ ] Documentation added
- [ ] Tests added
- [ ] E2e tests created

closes GH-390

### DIFF
--- a/libs/damap/src/assets/i18n/templates/en.json
+++ b/libs/damap/src/assets/i18n/templates/en.json
@@ -44,15 +44,18 @@
         },
         "contact": {
           "person": "Contact person",
+          "contributor": "Change to contributor",
           "add": "Set as contact",
-          "remove": "Remove as contact",
+          "remove": "Remove",
           "edit": "Edit contact person",
           "update": "Update",
-          "cancel": "Cancel"
+          "cancel": "Cancel",
+          "title": "Project contact",
+          "info": "Select a contact from the list of contributors using the button"
         },
         "manual.input": {
           "show": "Add contributor manually",
-          "hide": "Hide manual contributor input",
+          "hide": "Input contributor manually",
           "tooltip": "Use to add project contributor not found in database above",
           "question": {
             "firstname": "First name",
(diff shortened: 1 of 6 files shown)


GeoffreyKarnbach (changes requested):
Works great, nice work!

Would it be possible to add an action to the 
image
button to remove the current contributor as a contact person (and therefore go back to the blank contact person state, it is currently only possible by removing the contributor with the bin from the project, instead the contributor should be put back in the contributor list without being the contact person).

GeoffreyKarnbach (approved):
Everything works now as expected, great work!

PR #397, opened 2024-12-11
Ethical: Add new ethical report uri field to frontend  (disableable from the backend)

### LINKED ISSUE #382: Ethics: Add ethical issues report to DAMAP
**Description**
Below this part should be an inpuit field for the ethical issues report, see [RDA-Common-standard](https://github.com/RDA-DMP-Common/RDA-DMP-Common-Standard?tab=readme-ov-file#ethical_issues_report:~:text=yes-,ethical_issues_report,-To%20indicate%20where) for more info.
image

The field should be a valid URI.

The input field should be added, and only shown when yes was chosen in the question above. take care that the value in this field is only sent to the backend, when yes is ticked in the question above.

Before you start, see the [backend issue](https://github.com/tuwien-csd/damap-backend/issues/292) for more info.

GeoffreyKarnbach:
## Description

Feature

#### What does this PR do?

Add new ethical report URI field in the frontend step 6 ("Ethical aspects" view), the input field is only being displayed if it is not disabled through the config and if the Question "Was your research plan reviewed by an ethics committee, the Ethics Committee of your institution , or a similar body?" has been answered with yes.

A new URI and URL Validator have been introduced in a reusable manner.

#### Dependencies

[Related Backend PR](https://github.com/tuwien-csd/damap-backend/pull/302)

closes GH-382

### DIFF
--- a/apps/damap-frontend/src/app/services/config.service.spec.ts
+++ b/apps/damap-frontend/src/app/services/config.service.spec.ts
@@ -60,6 +60,7 @@ describe('ConfigService', () => {
         personSearchServiceConfigs: [],
         fitsServiceAvailable: false,
         livePreviewAvailable: true,
+        ethicalReportEnabled: true,
       };
 
       mockOAuthService.loadDiscoveryDocumentAndTryLogin.and.returnValue(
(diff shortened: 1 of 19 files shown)


ValentinFutterer:
Good work, the validators will also prove useful in the future!

Only one thing, like in the backend. Can you remove the mentions of URI and URL and rename them to ethical issues report like in the standard?

ValentinFutterer (approved):
Good work

PR #406, opened 2025-01-07
Storage: Reset fontsize to normal for dataset dropdown

### LINKED ISSUE #402: Step 5: Font in dropdown is too small
**Describe the bug**
The font in the dropdown is too small for the 
image

**Is this a regression?**
Yes.

**Expected behavior**
Font size should get smaller only when screen width diminishes.

**Environment**
- Damap version 4.3.1

GeoffreyKarnbach:
## Description

Bugfix

#### What does this PR do?

Fix font of text in dataset dropdown to regular instead of x-small

closes GH-402

### DIFF
--- a/libs/damap/src/lib/components/dmp/data-storage/storage/storage.component.css
+++ b/libs/damap/src/lib/components/dmp/data-storage/storage/storage.component.css
@@ -66,20 +66,11 @@
   }
 }
 
-::ng-deep .mat-mdc-option .mdc-list-item__primary-text {
-  white-space: normal;
-  word-wrap: break-word;
-  overflow-wrap: break-word;
-  padding: 10px;
-  font-size: small !important;
-}
-
 .mdc-menu-surface {
   max-width: 100%;
 }
 
-::ng-deep .mat-mdc-option .mdc-list-item__primary-text {
-  font-size: x-small !important;
+.mat-mdc-option .mdc-list-item__primary-text {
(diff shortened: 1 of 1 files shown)


libs/damap/src/lib/components/dmp/data-storage/storage/storage.component.css
```diff
@@ -66,20 +66,18 @@
   }
 }
 
-::ng-deep .mat-mdc-option .mdc-list-item__primary-text {
+.mat-mdc-option .mdc-list-item__primary-text {
   white-space: normal;
   word-wrap: break-word;
   overflow-wrap: break-word;
   padding: 10px;
```
ValentinFutterer: Removing the ::ng-deep here makes this piece of code not do anything. I am not against removing it, since ng-deep is deprecated, but I would like to know why it was introduced @lpandath . I see that it styles the multiple choice dropdown, but it only applies it to storages. Maybe this styling should be applied to all dropdowns? To replace ng-deep, we could use the [panelClass](https://material.angular.io/components/select/overview#adding-custom-styles-to-the-dropdown-panel).

lpandath: @ValentinFutterer to my knowledge, in the mobile version the text was going outside button container...

ValentinFutterer (approved):
Works

PR #407, opened 2025-01-07
Contributor: Fix diverse small UI bugs related to the editing of contributors

### LINKED ISSUE #401: Contributor bugs from rework
**Describe the bug**
image

Pressing edit opens two edit inputs - should obviously only open one. Also, there is not enough space between the hint sections of the inputs and the buttons below.  Same space problem also happens in the manual contributor input.

image

Tooltip has the wrong text, since this button removes the contact person as contact person anfd leaves the spot empty.
image

Here this edits a normal contributor and not the contact person.
image

This button removes the contributor as contact person, but the icon does not suggest that behaviour.
image

The icons are not inline with the input field. This is due to the mat-hint below the input field, but users dont know that, they only see that they are not aligned. I would suggest aligning the actual input field with the icons.
image

These are FAB buttons. FAb stands for floating action button, since they are supposed to trigger integral actions to the whole page, like saving. They are also supposed to float - meaning they should have a fixed position and be always accessible. Read more [here](https://m3.material.io/components/floating-action-button/overview). Therefore i would replace them with our standard buttons. Another reason to replace is, that this is the only place in the whole app where these kinds of buttons are used, so the styling is incoherent.

image

**Is this a regression?**
Yes.

**Expected behavior**
Only one edit input opens up. Location depends on if contact person or normal contributor was edited.

**Environment**

- Damap version 4.3.1

GeoffreyKarnbach:
## Description

Bugfix

#### What does this PR do?

See Issue for detailed description of fixed bugs

closes GH-401

### DIFF
--- a/libs/damap/src/assets/i18n/templates/en.json
+++ b/libs/damap/src/assets/i18n/templates/en.json
@@ -35,7 +35,8 @@
         "intro": "In this step, you add people involved in the project and provide additional info about them, e.g., their role. In this example, the project leader’s info has already been added and marked as the contact person (note icon with person and envelope next to their name). You can add more contributors via the search field (ensure you select the appropriate database, e.g., ORCID), or manually via the “Add contributor manually button”. Please provide a minimum of one contact person as well as a data manager (it can be the same person).",
         "search": "Search for person...",
         "search.service": "Search service",
-        "general.search": "Find person and service search solutions:",
+        "general.search": "Find Contributors",
+        "general.recommendations": "Recommended contributors:",
         "contributor": {
           "search": "Search contributor",
           "add": "Add as contributor",
@@ -46,8 +47,9 @@
           "person": "Contact person",
           "contributor": "Change to contributor",
           "add": "Set as contact",
-          "remove": "Remove",
+          "remove": "Remove contact person",
           "edit": "Edit contact person",
+          "editContributor": "Edit contributor",
           "update": "Update",
           "cancel": "Cancel",
           "title": "Project contact",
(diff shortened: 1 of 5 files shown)


ValentinFutterer:
Yes good work on addressing the issues 👍 
I noticed two more things.
First, the contributor has uneven amrgin isnide and (imo) unnecessary space to the top. I would remove this and center everything inside the card.

image

Secondly, the Search Contributor and Input Contributor Manually sections ahve different spacing to the top, this should be homogenized.

image

image

Regarding these two - when we have around 5 contributors, these sections start vanishing towards the borrom of the page and users have to scroll to reach them, after pressing a navigation button. Since these are the main functionality of the contributor section, I would move them to the top, if everyone agrees.

ValentinFutterer:
Another addition - it is not clear that those options here are supposed to be the recommendations based on project members. A header to make this clear would be good.

image

ValentinFutterer:
This title is very weird - it probbaly should just say: "Find Contributors", or something similar.
Write "Recommended contributors" instead of "Recommended people", to keep in line with the naming of the component.
image

When there are no project members, the recommended contributors part looks like this. I would remove the normal text below and change the title to "no recommended contributors available".
image

If there are no contributors assigned to a DMP, the contributor list is not shown, but the contact person reminder is. It should be hidden with the rest.
image

ValentinFutterer:
Additionally, i found that the formatting starts to break at 1250px:
image

andresTabiTuwien (approved):
All comments were fully addressed. Good work

PR #416, opened 2025-01-15
Global: make project title configurable via environment variable

### LINKED ISSUE #409: Project title configurable via env variable
**Changes**
Add the project title dynamically using environment variables. This way, we can set the title directly in the Docker Compose file for testing or production configurations. This way will allow each institution to have their DMP tool named according to their preference without requiring changes inside the project code.

**Additional notes**
List any notes you collected that support the original task and include a screenshot if applicable.

**Assumptions**
List any assumptions that you are aware of but still need to be reviewed.

**Acceptance Criteria**
Describe how you think a completed task would look like or how it should work.

**Risks? (optional)**
Potential risks and how to best mitigate them.

lpandath:
#### What does this PR do?

This PR adds support for dynamically setting the project title in the frontend application based on the backend-provided configuration. Previously, the title was statically set in the frontend. With this change, the title is dynamically fetched and applied from the backend’s /api/config endpoint.

#### Breaking changes

This PR does not introduce breaking changes. If the appTitle field is not provided by the backend, the application will default to “DAMAP Tool” as the browser title.

#### Code review focus

- Ensure the appTitle is correctly fetched from the backend and set as the browser title.
- Verify the frontend handles missing appTitle gracefully.
- Check the default title is used when appTitle is not provided.

Backend: https://github.com/tuwien-csd/damap-backend/issues/319

#### Steps

Dynamic Title Handling:

- Updated ConfigService to fetch and expose appTitle from /api/config.
- Added a getAppTitle method in ConfigService for dynamic title retrieval.
- Modified AppComponent to use ConfigService for setting the browser title.

Error Handling:

- Added fallback logic to use a default title if appTitle is missing.
- Logged a console warning when appTitle is not provided.

Unit Test Updates:

- Updated mock configurations to include appTitle.
- Added test cases for ConfigService to handle appTitle presence and fallback.
- Verified AppComponent dynamically sets the correct browser title.

### Checks

- [ ] Summary updated
- [ ] Version view updated
- [ ] Documentation added
- [ ] Tests added
- [ ] E2e tests created

closes GH-

### DIFF
--- a/libs/damap/src/lib/domain/config.ts
+++ b/libs/damap/src/lib/domain/config.ts
@@ -6,6 +6,7 @@ export interface Config {
   readonly authScope: string;
   readonly authUser: string;
   readonly env: string;
+  readonly appTitle: string;
   readonly personSearchServiceConfigs: ServiceConfig[];
   readonly fitsServiceAvailable: boolean;
   readonly livePreviewAvailable: boolean;
--- a/apps/damap-frontend/src/app/app.component.spec.ts
+++ b/apps/damap-frontend/src/app/app.component.spec.ts
@@ -2,16 +2,25 @@ import { TestBed, waitForAsync } from '@angular/core/testing';
 
 import { AppComponent } from './app.component';
 import { CUSTOM_ELEMENTS_SCHEMA } from '@angular/core';
+import { ConfigService } from './services/config.service';
 import { Title } from '@angular/platform-browser';
-import { environment } from '../environments/environment';
 
 describe('AppComponent', () => {
   let titleService: Title;
+  let mockConfigService: jasmine.SpyObj<ConfigService>;
+
   beforeEach(waitForAsync(() => {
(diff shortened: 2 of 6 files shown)


libs/damap/src/lib/domain/config.ts
GeoffreyKarnbach, line 1: You have changed the DO in the Frontend, but not in the Backend PR, is that intended?

GeoffreyKarnbach (changes requested):
The feature should work fine, but cannot be properly tested, as the matching backend PR does not contain the API changes. Is that intended?

libs/damap/src/lib/domain/config.ts
lpandath: @GeoffreyKarnbach its here: https://github.com/tuwien-csd/damap-backend/pull/320

GeoffreyKarnbach (approved):

PR #424, opened 2025-02-19
Dataset: Add technical resource for new datasets

### LINKED ISSUE #403: Dataset: Add technical resource input
**Description**
See the [backend issue](https://github.com/tuwien-csd/damap-backend/issues/295) for context. Users should be able to put in technical resources for new datasets, but not reused ones. A dataset can have multiple technical resources. A technical resource has a name and a description attribute, but only the name is required for our purposes, since this way its kept more userfriendly. (So you dont need the description input and null can be sent to the backend for description - I discussed this with Tomasz) Name input should be a textarea though, since user might want to put in multiple lines, but with a 255 character limit.

For wordig an dhelptext, consult the common standard and or Christiane/Tomasz.

Additionally, you have to solve the size problem here, the table is already dangerously small, especially with multiple technical resources:
image

Maybe add a menu point that shows additional info on demand, and only show the most relevant points per dataset in the table? Other ideas are also welcome.

GeoffreyKarnbach:
## Description

Feature

#### What does this PR do?

Add technical resource for new/produced datasets

closes GH-403

### DIFF
--- a/libs/damap/src/lib/components/dmp/specify-data/dataset-information/dataset-information.component.html
+++ b/libs/damap/src/lib/components/dmp/specify-data/dataset-information/dataset-information.component.html
@@ -1,0 +1,52 @@
+<h1 mat-dialog-title>
+  {{ dataset.title }}
+</h1>
+
+<mat-dialog-content id="datasetDialog">
+  <div>
+    <div class="info-section">
+      <span class="label"
+        >{{ "dmp.steps.data.specify.dialog.type" | translate }}:</span
+      >
+      <span class="split-text">{{ getDataType(dataset) }}</span>
+    </div>
+
+    <div class="info-section">
+      <span class="label"
+        >{{ "dmp.steps.data.specify.dialog.fileFormat" | translate }}:</span
+      >
+      <span class="split-text">{{ dataset.fileFormat }}</span>
+    </div>
+
+    <div class="info-section">
+      <span class="label"
+        >{{ "dmp.steps.data.specify.dialog.size" | translate }}:
+      </span>
+      <span>{{ mapFileSize(dataset.size) }}</span>
+    </div>
+  </div>
+
+  <div class="description-section">
+    <span class="label"
+      >{{ "dmp.steps.data.specify.dialog.description" | translate }}:
+    </span>
+    <p class="split-text">{{ dataset.description }}</p>
+  </div>
+
+  <div class="technical-resources">
+    <div
+      class="resource-box"
+      *ngFor="let resource of dataset.technicalResources; let i = index">
+      <span class="resource-title">{{
+        "Technical Resource " + (i + 1) + ":"
+      }}</span>
+      <p class="split-text">{{ resource.name }}</p>
+    </div>
+  </div>
+</mat-dialog-content>
+
+<mat-dialog-actions>
+  <button mat-button (click)="onDialogClose()">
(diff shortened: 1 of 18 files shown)


libs/damap/src/lib/components/dmp/specify-data/dataset-information/dataset-information.component.html
```diff
@@ -0,0 +1,90 @@
...
+  .dataset-info,
+  .description-section,
+  .technical-resources {
+    margin-bottom: 16px;
+  }
+
+  .info-section {
+    display: flex;
+    justify-content: space-between;
+    background: #f5f5f5;
```
ValentinFutterer: You should use colors from our color palette here.

libs/damap/src/lib/components/dmp/specify-data/dataset-information/dataset-information.component.html
```diff
@@ -0,0 +1,90 @@
...
+    border-radius: 4px;
+    margin-bottom: 8px;
+  }
+
+  .label {
+    font-weight: bold;
+  }
+
+  .resource-box {
+    background: #f5f5f5;
```
ValentinFutterer: You should use colors from our color palette here.

ValentinFutterer:
Great work on the interface for the technical resource and especially the popup for viewing the whole dataset - very nice addition to the project 💪 

I tried out a few edge cases (like usual) and want to share my results:

You could add a little space here for all lines and keep them seperated.
image

Estimated size shows a number, but in the sizepicker, you choose a size range. There is an issue to fix that in the DB and make everything into ranges, but for know we should keep the implemenation like is i think. Can you refactor it and show the ranges here?
image

And i know this wasnt in the scope of the issue, but the table had problems with longer entries and still has. The new dataset popup view helps, but we still ned to fix the table itself. I would suggest to first redistribute the space a little and then to cut long text using `...` when a certain length is reached.

image

image

And something else in the popup view, if you have a long text without spaces it reaches out of the border.
image

ValentinFutterer:
Ah and description doesnt have a box like the other fields, was this on purpose?
image

libs/damap/src/lib/components/dmp/specify-data/dataset-information/dataset-information.component.html
GeoffreyKarnbach: done

libs/damap/src/lib/components/dmp/specify-data/dataset-information/dataset-information.component.html
GeoffreyKarnbach: done

ValentinFutterer:
Estimated type in the dataset view still spills over:

image

I think there could be an argument made for either removing the title `View Dataset`, or to put the name of the dataset next to the title. I think it looks a little weird right now, since they are not aligned and the name is bold, while the title is not.

image


There are irregular distances between the elements, this should be rectified.
image

Datatype isnt shortened by the ... when it gets too long, which is a very edge case, but while we are at it, we can add it

image

I think the space distribution in the columns is off - description should be preferred over file format, since it will be a longer text. Otherwise this can happen: We want to avoid the table looking like this. Maybe the shortening with ... should happen earlier?
image

GeoffreyKarnbach (commented):
LGTM!

ValentinFutterer (approved):
Rdy to be merged

PR #431, opened 2025-03-25
DMP table: Fix pagination

### LINKED ISSUE #420: DMPs Paginator is still bugged
**Describe the bug**

The DMP paginator in the plans sections is bugged on certain devices (seems independent of OS). It just shows 0/0  and doesnt paginate at all. Someone on who's device the paginator doesnt work needs to fix this, since it seems that the bug also occurs on some production instances.

Image

Also make sure that the sorting on top of the paginator is working as expected:

Image

**Is this a regression?**
Yes.

**To Reproduce**
Unknown - seems to randomly work on some devices and not on others. But consistently fails on the same device.

**Expected behavior**
Paginator should work as expected

**Environment**

DAMAP 4.3.1

ValentinFutterer:
## Description

Bugfix

#### What does this PR do?

Fix the paginator by moving the "if DMPs are loaded" code. It would prevent the paginator from being loaded in the typescript code, since it was not poresent in the DOM due to the *ngIf. Also fixed other small errors.

closes GH-420

### DIFF
--- a/libs/damap/src/lib/widgets/dmp-table/dmp-table.component.html
+++ b/libs/damap/src/lib/widgets/dmp-table/dmp-table.component.html
@@ -11,7 +11,7 @@
     (searchChange)="applyFilter($event)"></app-search-field>
 </div>
 
-<div class="table-wrapper" *ngIf="(dmpsLoaded | async) === LoadingState.LOADED">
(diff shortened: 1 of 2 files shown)


GeoffreyKarnbach (approved):
Works great, nice catch!

PR #439, opened 2025-04-07
Costs: Change stepSize to 100

### LINKED ISSUE #384: Costs: Increment is too small
**Changes**
Raising the costs with the arrows works in cent increments:
image

This is obviously not suitable for real costs, a better increment should be chosen here. If memory serves, this increment is the base one in our standardised textfield component, the change should be simple. Med Uni Graz is going to give us some pointers on how good increments should look like. Christiane can also be queried about this.

**Acceptance Criteria**
New, fitting increment has been implemented.

BlueTechie:
## Description
Refactor

#### What does this PR do?
Changed the increment/decrement button to 100 instead of 0.01 when adding a new cost via stepSize input

389503722-b6017326-755a-420b-837c-b3b59b5113ec

#### Dependencies

closes GH-384

### DIFF
--- a/libs/damap/src/lib/components/dmp/costs/costs.component.html
+++ b/libs/damap/src/lib/components/dmp/costs/costs.component.html
@@ -69,6 +69,7 @@
                     [label]="'dmp.steps.costs.question.value'"
                     [prefix]="'€ &nbsp;'"
                     [placeholder]="'1000.00'"
+                    [stepSize]="100"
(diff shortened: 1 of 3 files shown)


ValentinFutterer (approved):
Good work adding the `@Input`
As discussed, we will leave the rounding functionality of the stepper as is, since its the default behaviour of the component.

PR #441, opened 2025-04-07
Global: make no datasets message more clear for step 7/8

### LINKED ISSUE #393: Global: Make no datasets message more clear for step 7/8
**Describe the bug**
When no dataset is declared, step 3-10 are locked behind this message:
image

This message can be unclear though, since step 7-8 need a new dataset specifically. When only a reused dataset is declared, only these two steps remain locked, but with the unclear error message. For these two steps, it should say `Please declare a produced dataset first`. If you want to be fancy, you could link this back to the new dataset tab. The same could be done for the other error step messages.

**Is this a regression?**
No.

**To Reproduce**
Create a dmp and an existing dataset.

**Expected behavior**
Clear error message + linking for convenience.

**Environment**

- Damap version 4.3.0

BlueTechie:
## Description
UX Improvement

#### What does this PR do?
Clarified the error message for steps 7/8 to notify the user that a new dataset must be created first to unlock those steps.
Clicking these error messages (and the ones where any dataset must be defined) routes to step 3 (specify research data).

closes GH-393

### DIFF
--- a/libs/damap/src/lib/components/dmp/dmp.component.html
+++ b/libs/damap/src/lib/components/dmp/dmp.component.html
@@ -298,7 +298,7 @@
       </ng-container>
     </ng-container>
     <ng-container *ngIf="selectedStep === 6">
-      <ng-container *ngIf="showStepIfNewDatasets; else noDatasets">
+      <ng-container *ngIf="showStepIfNewDatasets; else noNewDataset">
         <app-dmp-licenses
           [dmpForm]="dmpForm"
           [datasets]="datasets"></app-dmp-licenses>
@@ -308,7 +308,7 @@
       </ng-container>
     </ng-container>
     <ng-container *ngIf="selectedStep === 7">
-      <ng-container *ngIf="showStepIfNewDatasets; else noDatasets">
+      <ng-container *ngIf="showStepIfNewDatasets; else noNewDataset">
         <app-dmp-repo
           #repo
           [dmpForm]="dmpForm"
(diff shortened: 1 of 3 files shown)


ValentinFutterer:
Works well, but the linked text looks like normal text - there is no way for a user to know that this is clickable. I would suggest to either add a navigation button or to style the text like a link (blue and underscore).
image

Also sidenote: I talked with Christiane and it would make sense to also have step 9 locked when there is no new dataset. Could you please add this, while you are at it?

libs/damap/src/lib/components/dmp/dmp.component.html
```diff
@@ -352,7 +352,15 @@
...
-  {{ "dmp.nodatasets" | translate }}
+  <div class="no-dataset-text" (click)="onNewDatasetClick()">
+    {{ "dmp.nodatasets" | translate }}
+  </div>
+</ng-template>
+
+<ng-template #noNewDataset>
+  <div class="no-new-dataset-text" (click)="onNewDatasetClick()">
+    {{ "dmp.nonewdataset" | translate }}
+  </div>
```
ValentinFutterer: You could replace this with:
```
<ng-template #noDatasets>
  <a href="javascript:void(0)" (click)="onNewDatasetClick()">
    {{ "dmp.nodatasets" | translate }}
  </a>
</ng-template>

<ng-template #noNewDataset>
  <a href="javascript:void(0)" (click)="onNewDatasetClick()">
    {{ "dmp.nonewdataset" | translate }}
  </a>
</ng-template>
```

Then you dont need a custom styling css class and can just use the inbuilt one.
Or you create a global link style and use this class.
Or you just use a button to navigate.

BlueTechie: I have incorporated the changes 👍

ValentinFutterer (approved):
Looks good now 👍

PR #442, opened 2025-04-08
Manual Project: Make Update Project button more user friendly (feedba…

### LINKED ISSUE #383: Manual Project: Make button more user friendly
**Changes**
The `update project` button is not very user friendly - to save a manually added project, users have to first click on it and then on save. Clicking on save first does nothing. There is also no feedback that clicking the `update project` button did anything. At the very least, there should be some indication that pressing this button did something. A better solution would maybe be to save automatically after the `update project` button was pressed.

image

**Acceptance Criteria**
There is some feedback for the `update project` button at least. Some other, bigger reworks are also acceptable. Be creative, maybe you have a good idea.

BlueTechie:
## Description

Feature
#### What does this PR do?
Made "update project" button more user friendly: 
- Button reads "Update project info" if it is meant to update the info of an already existing project, "Add Project" if no selected project exists
- Fixed some labels where the user inputs the project information manually for clarity
- User now receives feedback when the project information has been updated (in case of only changing the description for instance, before no feedback was given)

closes GH-383

### DIFF
--- a/libs/damap/src/assets/i18n/templates/en.json
+++ b/libs/damap/src/assets/i18n/templates/en.json
@@ -2,12 +2,17 @@
   "dmp": {
     "steps": {
       "project": {
+        "success": "Updated project information!",
         "input": {
           "title": "Project title",
           "description": "Project description",
-          "button": "Update project",
-          "acronym": "Update acronym",
-          "date": "Enter start and end date",
+          "info": "Project information",
+          "button": {
+            "existing": "Update project info",
+            "new": "Add project"
+          },
+          "acronym": "Acronym",
+          "date": "Start and end date",
           "datehint": "DD/MM/YYYY – DD/MM/YYYY"
         },
         "manual.input": "Input project manually",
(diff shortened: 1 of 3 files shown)


ValentinFutterer:
There is still some basic stuff going wrong -  clicking on update Project doesnt save the DMP for example, you need to click twice.
Another fishy thing that happens is when you double click to save, then delete the project and click around three times on the update button, you get an error.

The other question is, if its even necessary to save at this point - I think its enough if you orient yourself on the behaviour of the choose project step and mimic its behaviour.
When you add a contributor manually, it also does not save, for example.

An additional note: "Update Project" doesnt really make sense when there is no project yet. You could make the button say "Add project", when there is no chosen project.

BlueTechie:
Ah yes this makes sense, I have updated the PR description and in particular made it so manually adding or updating a project behaves like the "Choose project" step.

ValentinFutterer (approved):
Looks good to me

PR #443, opened 2025-04-09
Storage: Add checkbox to indicate if an external storage is managed

### LINKED ISSUE #437: Add option to distinguish external storage from "private" institutional services
**Description**
Some institutions have private storage options, that are not none to maintainers of DAMAP instances, and are therefore not in recommended storages, nor could they be, since they are only recommended for certain institutions. When they pick other storages, read external storages, the export document lists them as external, which is clearly wrong, We should therefore have an option to distinguish the two cases from each other. The `Please explain why institutional storage should not be used option` should of course be hidden, when the non external option has been chosen.

option to distinguish between the two cases, plus hidden text field. Also an explanation on what the slider or box does.

**Additional context**
backend issue: https://github.com/tuwien-csd/damap-backend/issues/349

GeoffreyKarnbach:
## Description

Feature

closes GH-437

### DIFF
--- a/libs/damap/src/lib/components/dmp/data-storage/external-storage/external-storage.component.html
+++ b/libs/damap/src/lib/components/dmp/data-storage/external-storage/external-storage.component.html
@@ -73,9 +73,20 @@
             </div>
           </div>
         </mat-card-content>
+        <mat-card-actions>
+          <mat-slide-toggle
+            [formControl]="getFormControl(i, 'isManagedInternally')">
+            {{
+              "dmp.steps.storage.external.question.isManagedInternally"
+                | translate
+            }}
+          </mat-slide-toggle>
+        </mat-card-actions>
       </mat-card>
     </ng-container>
-    <div *ngIf="externalStorageStep?.controls.length" class="storage-textarea">
+    <div
+      *ngIf="externalStorageStep?.controls.length && anyNonInternalStorage()"
+      class="storage-textarea">
       <app-textarea-wrapper
         [label]="'dmp.steps.storage.external.question.info'"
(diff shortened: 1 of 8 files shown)


libs/damap/src/lib/components/dmp/data-storage/external-storage/external-storage.component.html
```diff
@@ -72,10 +72,19 @@
                 "></app-input-wrapper>
             </div>
           </div>
+          <div class="form-row">
+            <mat-slide-toggle
+              [checked]="getFormControl(i, 'isInternal').value"
+              (change)="toggleInternalStorage(i)">
+              {{ "dmp.steps.storage.external.question.isInternal" | translate }}
+            </mat-slide-toggle>
```
ValentinFutterer: Is there a reason why you are not just binding this to the formcontrol?

ValentinFutterer:
Have you checked with Livia and Christiane about the wording here? I am not sure that most users will understand what this button does.
image

You should also try to play around with mat-card action / footer sections - buttons like this are usually put inside of them (different styling).

Same as in the Backend PR, I would try to find a different name for isInternal. Its going to be really confusing to devs what it does and why it needs to exist when we already have an InternalStorage class.

ValentinFutterer:
> Have you checked with Livia and Christiane about the wording here? I am not sure that most users will understand what this button does.

Seems like you didnt check with them, I did it for you. Christiane was in favor of this wording:
`Is this storage provided by a department or service unit of your institution?`

ValentinFutterer (approved):
Ready to merge 👍

PR #444, opened 2025-04-09
Fix: recommended contributors panel behavior

### LINKED ISSUE #432: Fix small issues in recommended contributor section
**Describe the bug**
The recommended contributor expansion panel always closes when you add a contributor from the panel - this is annoying since users need to open - add, open - add, open - add...

The panel should only autoclose when there are no recommended contributors. We could also go a step further and just hide the whole thing in this case.

Image

Bonus: Think about adding an AddAll button, if it fits in the design.

**Is this a regression?**
No.

**Expected behavior**
Should not autoclose when adding a contributor and should handle the no recommended contributor available step.

**Environment**

- Damap version 4.4.0

lpandath:
## Description

https://github.com/user-attachments/assets/4f1759a8-93a4-4617-936f-39eea0125b4f

Screenshot 2025-05-20 at 08 45 59

#### What does this PR do?

- Panel stays open when adding contributors
- Only closes when no recommendations left
- Add "Add All" button for bulk adding contributors

### Checks

- [ ] Summary updated
- [ ] Version view updated
- [ ] Documentation added
- [x] Tests added
- [ ] E2e tests created

closes GH-432

### DIFF
--- a/libs/damap/src/lib/components/dmp/people/people.component.html
+++ b/libs/damap/src/lib/components/dmp/people/people.component.html
@@ -47,7 +47,11 @@
   </mat-label>
 
   <mat-expansion-panel
-    *ngIf="projectMembers?.length > 0"
+    *ngIf="
+      projectMembers?.length > 0 &&
+      (projectMembers | contributorFilter: dmpForm.controls.contributors.value)
+        .length > 0
+    "
     [expanded]="!isCollapsed"
     (opened)="isCollapsed = false"
     class="white-background-panel">
@@ -56,6 +60,20 @@
         {{ "dmp.steps.people.general.recommendations" | translate }}
       </mat-panel-title>
     </mat-expansion-panel-header>
+
+    <div
+      class="add-all-container"
+      *ngIf="
+        (
+          projectMembers
+          | contributorFilter: dmpForm.controls.contributors.value
+        ).length > 0
+      ">
+      <button mat-button color="primary" (click)="addAllContributors()">
+        <mat-icon>group_add</mat-icon>
+        {{ "dmp.steps.people.general.addAll" | translate }}
+      </button>
+    </div>
 
     <div class="list-container">
       <mat-card
@@ -70,7 +88,7 @@
               <mat-icon class="mat-icon-style">person</mat-icon>
               {{ member.firstName }} {{ member.lastName }}
             </div>
-            <mat-icon class="mat-icon-style outline">mail</mat-icon>
+            <mat-icon class="mat-icon-style outline">mail_outline</mat-icon>
             {{ member.mbox }}
           </div>
           <div class="mat-card-right">
@@ -123,14 +141,16 @@
             <div class="mat-card-container">
               <div class="mat-card-left">
                 <div>
-                  <mat-icon class="mat-icon-style">person</mat-icon>
+                  <mat-icon class="mat-icon-style outline"
+                    >person_outline</mat-icon
+                  >
                   {{ contributor.value.firstName }}
                   {{ contributor.value.lastName }}
                   <span class="sr-only">{{
                     "dmp.steps.people.contact.person" | translate
                   }}</span>
                 </div>
-                <mat-icon class="mat-icon-style outline">mail</mat-icon>
+                <mat-icon class="mat-icon-style outline">mail_outline</mat-icon>
                 {{ contributor.value.mbox }}
                 <app-orcid
                   *ngIf="
@@ -224,7 +244,7 @@
   </ng-container>
 </div>
 
-<div *ngIf="contactContributor() === -1 && contributors.length !== 0">
+<div *ngIf="!doesContactExist() && contributors.length !== 0">
   <mat-card appearance="raised" class="card-selected-empty">
     <mat-card-content>
       <mat-icon class="material-icon-primary outline">info</mat-icon>
@@ -253,7 +273,9 @@
             <div class="mat-card-container">
               <div class="mat-card-left">
                 <div>
-                  <mat-icon class="mat-icon-style">person</mat-icon>
+                  <mat-icon class="mat-icon-style outline"
+                    >person_outline</mat-icon
+                  >
                   {{ contributor.value.firstName }}
                   {{ contributor.value.lastName }}
                   <mat-icon
@@ -265,7 +287,7 @@
                     "dmp.steps.people.contact.person" | translate
                   }}</span>
                 </div>
-                <mat-icon class="mat-icon-style outline">mail</mat-icon>
+                <mat-icon class="mat-icon-style outline">mail_outline</mat-icon>
                 {{ contributor.value.mbox }}
                 <app-orcid
                   *ngIf="
--- a/libs/damap/src/lib/components/dmp/people/people.component.html
+++ b/libs/damap/src/lib/components/dmp/people/people.component.html
@@ -141,9 +141,7 @@
             <div class="mat-card-container">
               <div class="mat-card-left">
                 <div>
-                  <mat-icon class="mat-icon-style outline"
-                    >person_outline</mat-icon
-                  >
+                  <mat-icon class="mat-icon-style">person</mat-icon>
                   {{ contributor.value.firstName }}
                   {{ contributor.value.lastName }}
                   <span class="sr-only">{{
@@ -273,9 +271,7 @@
             <div class="mat-card-container">
               <div class="mat-card-left">
                 <div>
-                  <mat-icon class="mat-icon-style outline"
-                    >person_outline</mat-icon
-                  >
+                  <mat-icon class="mat-icon-style">person</mat-icon>
                   {{ contributor.value.firstName }}
                   {{ contributor.value.lastName }}
                   <mat-icon
--- a/libs/damap/src/lib/components/dmp/people/people.component.ts
+++ b/libs/damap/src/lib/components/dmp/people/people.component.ts
@@ -126,51 +126,11 @@ export class PeopleComponent implements OnInit, OnDestroy {
   }
 
   changeContactPerson(contact: Contributor): void {
-    // fetch accurate affiliation information from ORCID
-    // this should probably be done in the backend, with newly added contributors only
-    // since the affiliation is not displayed in the frontend anyways
-    if (contact.personId.type === IdentifierType.ORCID) {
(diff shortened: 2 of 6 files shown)


libs/damap/src/lib/components/dmp/people/people.component.html
```diff
@@ -55,15 +55,20 @@
...
+      </button>
     </mat-expansion-panel-header>
 
     <div class="list-container">
       <mat-card
         appearance="raised"
-        *ngFor="
-          let member of projectMembers
-            | contributorFilter: dmpForm.controls.contributors.value
-        ">
```
ValentinFutterer: Is there any reason you replaced this pipe with the `getFilteredProjectMembers` function? As i can see, both do essentially the same thing and the pipe is still in use in another place.

ValentinFutterer:
Works pretty well all in all.

Right now only the add all button is hidden, but imo we could just hide the whole panel. If you have a project with no contributors, the empty panel could be confusing. Same, when you added all recommended contributors, the panel serves no value then. What do you think?

Also minor bug: When you add a contact person, the panel still closes.

libs/damap/src/lib/components/dmp/people/people.component.ts
```diff
@@ -215,6 +225,49 @@ export class PeopleComponent implements OnInit, OnDestroy {
...
+  }
+
+  addAllContributors(): void {
+    const filteredMembers = this.getFilteredProjectMembers();
+
+    filteredMembers.forEach(member => {
+      this.contributorToAdd.emit(member);
+    });
+
+    this.contactContributor();
```
ValentinFutterer: Suggestion for refactoring: All usages of this method in the ts can be safely removed, its return values are never used. Its only being used in the html.
The method should also return a boolean and be named soemthing like: does contributor exist.

libs/damap/src/lib/components/dmp/people/people.component.ts
```diff
@@ -215,6 +225,49 @@ export class PeopleComponent implements OnInit, OnDestroy {
...
+
+        if (member.mbox && existing.mbox) {
+          return member.mbox === existing.mbox;
+        }
+        return false;
+      });
+
+      return !isAlreadyAdded;
+    });
+  }
```
ValentinFutterer: Writing function code is not necessary, it duplicates functionality from a pipe and the contributor compare function, you could just replace this with a call to the pipe function.

libs/damap/src/lib/components/dmp/people/people.component.ts
ValentinFutterer: For clarification: i am talking about `contactContributor()`.

libs/damap/src/lib/components/dmp/people/people.component.html
lpandath: the function has some extra null checks and handles the form access directly

libs/damap/src/lib/components/dmp/people/people.component.ts
lpandath: Thank you for the suggestion. You're right that the getFilteredProjectMembers() method duplicates functionality that already exists in the ContributorFilterPipe.

libs/damap/src/lib/components/dmp/people/people.component.html
ValentinFutterer:
```
if (!this.projectMembers || !this.dmpForm) {
      return [];
    }
```

Does this guard against something thats realistically going to happen? Because to my knowledge, the pipe has never failed before.
And if this is a realistic danger, shouldnt the pipe have the same safeguards, since the pipe is still in use in other places?

I think best practice would be to write a function that does the filtering and call it from the pipe and  from the `getFilteredProjectMembers` function, so you can avoid having to call the pipe with transform.
But it would be easier to just keep it as is and use transform in `getFilteredProjectMembers`, I guess. 

I would just move the guards (if necessary) into the pipe and use the actual pipe in the html files and the wrapper that calls the pipes in ts files, to keep them seperate and to make it consistent, since the pipe is still in use.

ValentinFutterer:
> Right now only the add all button is hidden, but imo we could just hide the whole panel. If you have a project with no contributors, the empty panel could be confusing. Same, when you added all recommended contributors, the panel serves no value then. What do you think?
> 
> Also minor bug: When you add a contact person, the panel still closes.

Could you address this?

ValentinFutterer:
Something else I noticed while clicking through it, the clickable surface of the "Add All" button and the closing button for the tab actually overlap.
image

I clicked the add all button by accident, when I tried it out. A good alternative would be to move the button to here, in a fixed position, so its always visible when scrolling. Or you just add it below the scrollable container:
image

libs/damap/src/lib/components/dmp/people/people.component.html
lpandath: Well in this case i would imp an shared filtering function because the pipe imp it only has a null check for newContributors (using the optional chaining operator ?.) but doesn't check if alreadyAddedContributors is null...

lpandath:
> > Right now only the add all button is hidden, but imo we could just hide the whole panel. If you have a project with no contributors, the empty panel could be confusing. Same, when you added all recommended contributors, the panel serves no value then. What do you think?
> > Also minor bug: When you add a contact person, the panel still closes.
> 
> Could you address this?

yes

libs/damap/src/lib/components/dmp/people/people.component.html
ValentinFutterer: Yes i think thats the best and most clean approach 👍 
I think you can just copy the code from the pipe to do this.

lpandath: no, i mean extending the ContributorFilterPipe, not copying it!

libs/damap/src/lib/components/dmp/people/people.component.html
```diff
@@ -123,14 +125,16 @@
             <div class="mat-card-container">
               <div class="mat-card-left">
                 <div>
-                  <mat-icon class="mat-icon-style">person</mat-icon>
+                  <mat-icon class="mat-icon-style outline"
+                    >person_outline</mat-icon
+                  >
```
ValentinFutterer: Is there a reason to change this to an outline icon. It kindof goes under due to the grey background. The mail is way bulkier and sticks out more, so it works better for that one.

lpandath: I wanted to have that on click, it's just a preference.

libs/damap/src/lib/components/dmp/people/people.component.ts
```diff
@@ -215,13 +216,35 @@ export class PeopleComponent implements OnInit, OnDestroy {
...
   }
 
-  contactContributor(): number {
-    let contributors = this.dmpForm.get('contributors') as UntypedFormArray;
-    return contributors.controls.findIndex(
-      (contributor, index) => contributor.value.contact,
+  private getFilteredMembers(): Contributor[] {
+    if (!this.projectMembers || !this.dmpForm) {
+      return [];
+    }
```
ValentinFutterer: You moved the check into the filterContributors function, so this can be safely removed.

libs/damap/src/lib/components/dmp/people/people.component.html
ValentinFutterer: What do you mean with on click?

libs/damap/src/lib/components/dmp/people/people.component.html
ValentinFutterer: Ah i was referring to the shared function, as in copy the pipe code into the shared function :)

ValentinFutterer:
> Also minor bug: When you add a contact person, the panel still closes.
> Could you address this?

This bug is still there and I mentioned this twice already. I also noticed that the pipe is still not used in the html file. Thats the sole purpose of pipes and thats also why we still have the pipe / why it was originally coded as a pipe and not as a function. We only need the `getFilteredProjectMembers` function for the ts file, since its not good practice to use pipes in typescript.

This is the third time I reviewed it and in this timeframe I could have basically finished this issue on my own. I try to do my reviews very thoroughly and its my biggest timesink by far. If you are stressed and dont have enough time then please wait until you have enough time to finish it or ask for help.

lpandath:
> > Also minor bug: When you add a contact person, the panel still closes.
> > Could you address this?
> 
> This bug is still there and I mentioned this twice already. I also noticed that the pipe is still not used in the html file. Thats the sole purpose of pipes and thats also why we still have the pipe / why it was originally coded as a pipe and not as a function. We only need the `getFilteredProjectMembers` function for the ts file, since its not good practice to use pipes in typescript.
> 
> This is the third time I reviewed it and in this timeframe I could have basically finished this issue on my own. I try to do my reviews very thoroughly and its my biggest timesink by far. If you are stressed and dont have enough time then please wait until you have enough time to finish it or ask for help.

Not sure which bug you mean: this here written:
```
Right now only the add all button is hidden, but imo we could just hide the whole panel. If you have a project with no contributors, the empty panel could be confusing. Same, when you added all recommended contributors, the panel serves no value then. What do you think?
Also minor bug: When you add a contact person, the panel still closes.
```
is resolved! 
can you please properly test it before commenting, would appriciate that. Thank you!

https://github.com/user-attachments/assets/9aeb1f4e-da3b-4935-a9bc-8a1d77995286

libs/damap/src/lib/components/dmp/people/people.component.html
```diff
@@ -123,14 +141,16 @@
...
```
ValentinFutterer: Can you please revert this? it doesnt look as good on the white/gray background and some of those icons at another place  are still black - so we have an inconsistency .
```suggestion
                  <mat-icon class="mat-icon-style">person</mat-icon>
```

libs/damap/src/lib/components/dmp/people/people.component.html
```diff
@@ -253,7 +273,9 @@
...
```
ValentinFutterer: Can you please revert this? it doesnt look as good on the white/gray background and some of those icons at another place are still black - so we have an inconsistency .
```suggestion
                  <mat-icon class="mat-icon-style">person</mat-icon>
```

ValentinFutterer (approved):
We can merge

PR #459, opened 2025-05-20
Stepper: Fix inaccuracies and oddities

### LINKED ISSUE #446: Stepper icon inaccuracies and oddities
> New Stepper icons also do not show as locked in the ase described here - this should also be fixed. 

 _Originally posted by @ValentinFutterer in [#393](https://github.com/tuwien-csd/damap-frontend/issues/393#issuecomment-2661667375)_

The new stepper mostly works fine, but there are a few edgecases where it doesnt function correctly.

When i add a dataset, the stepper is blank until it is saved:

Image

Step 7,8 (and 9 when the above issue is merged) should be locked when only an existing dataset is added, but they present like they are unlocked:

Step 5 is never shown as half-done, only empty until a storage is assigned to a dataset. (Also tied to the message in the last step)

Deleting all datasets leads to this for some reaso :) - until saved, then they turn normal.

Image

There could be more, try around yourself and see if you find anything.

BlueTechie:
## Description
Bugfix

#### What does this PR do?
Fixed the bugs specified in the issue below. 

For the first one (some icons being updated only after changes are saved, not after deleting a dataset), this seems to happen because of another issue with the summary service not updating correctly when deleting a dataset and will be fixed in a future PR.

Added and improved the clarity of some status messages in the summary service. 

Set the completeness of steps 7,8 & 9 to 0 if no new dataset has been specified (more precisely if no open new dataset for step 8 / no restricted new dataset for step 9 has been specified).

closes GH-446

### DIFF
--- a/libs/damap/src/assets/i18n/templates/en.json
+++ b/libs/damap/src/assets/i18n/templates/en.json
@@ -450,18 +450,19 @@
           }
         },
         "storage": {
-          "nonestored": "No (produced) datasets to be stored.",
-          "nodata": "No data defined yet.",
-          "alldata": "All (produced) datasets are stored.",
+          "allinfo": "All information necessary provided.",
+          "nonestored": "No datasets will be stored, as no produced or reused datasets have been specified.",
+          "missingstorage": "Some data is not stored.",
           "missingexplanation": " Usage explanation is missing."
         },
         "licensing": {
           "complete": "All information necessary provided.",
-          "incomplete": "Some information is missing."
+          "incomplete": "Some information is missing.",
+          "noproduceddata": "No produced dataset has been specified."
         },
         "repositories": {
           "nodata": "No data to be deposited.",
-          "nodataspecified": "No data defined yet.",
+          "noproduceddata": "No produced dataset has been specified.",
           "deposited": {
             "incomplete": "Not all data deposited yet.",
             "complete": "All data are deposited."
--- a/libs/damap/src/assets/i18n/templates/en.json
+++ b/libs/damap/src/assets/i18n/templates/en.json
@@ -453,7 +453,7 @@
           "allinfo": "All information necessary provided.",
           "nonestored": "No datasets will be stored, as no produced or reused datasets have been specified.",
           "missingstorage": "Some data is not stored.",
-          "missingexplanation": " Usage explanation is missing."
+          "missingexplanation": "Explanation for using external storage is missing."
         },
         "licensing": {
           "complete": "All information necessary provided.",
--- a/libs/damap/src/lib/components/dmp/dmp.component.ts
+++ b/libs/damap/src/lib/components/dmp/dmp.component.ts
@@ -203,7 +203,7 @@ export class DmpComponent implements OnInit, OnDestroy {
 
   get showStepIfNewDatasets() {
     return (
-      this.specifyDataStep.value.kind === DataKind.SPECIFY &&
+      this.specifyDataStep?.value.kind === DataKind.SPECIFY &&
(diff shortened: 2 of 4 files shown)


GeoffreyKarnbach (approved):
LGTM!

bitmotivator (approved):

libs/damap/src/assets/i18n/templates/en.json
```diff
@@ -444,18 +445,19 @@
...
         "licensing": {
           "complete": "All information necessary provided.",
-          "incomplete": "Some information is missing."
+          "incomplete": "Some information is missing.",
+          "noproduceddata": "No produced dataset has been specified."
         },
         "repositories": {
           "nodata": "No data to be deposited.",
-          "nodataspecified": "No data defined yet.",
+          "nonewdataset": "No produced dataset with open access has been specified.",
```
ValentinFutterer: This is a common misconception that only open data should be stored in repositories. restricted and even closed data should also be kept in repositories. This way, metadata about the data is still accessible. In the case for restricted data, it makes it easy to publish access conditions, since some people might have a legitimate reason to access restricted datasets. Repositories are also a much safer place to store data than a random harddrive (if you trust them)

libs/damap/src/assets/i18n/templates/en.json
```diff
@@ -444,18 +445,19 @@
           }
         },
         "storage": {
-          "nonestored": "No (produced) datasets to be stored.",
+          "nonestored": "No datasets to be stored since no produced dataset has been specified.",
```
ValentinFutterer: Its true that storage is mostly used for produced datasets, but it can also be relevant for reused datasets. Imagine you work with sensitive data - reviewers of the DMP might have an interest in knowing where you store this data during your project.

libs/damap/src/lib/components/dmp/dmp.component.ts
```diff
@@ -407,85 +408,94 @@ export class DmpComponent implements OnInit, OnDestroy {
...
+    style: string = '',
   ): boolean {
+    const inFirstRange = index < 3;
+    const inSecondRange = index >= 3 && index < 6;
+    const inThirdRange = index >= 6 && index < 9;
+    const isCosts = index === 9;
+    const isSummary = index === 10;
+
+    const completeness = this.checkCompletenessForm();
+
```
ValentinFutterer: I get wanting to group the steps, but using ranges is the wrong approach imo.  If you want to keep this, I would make it more verbose what these ranges represent by renaming. I like the isCosts, since you can expand it easily.

ValentinFutterer:
Trying to fix errors in the summary component is good, but I noticed some stuff thats not working correctly.

Why does this message show? I filled out everything in step 9? Why would I need a new restricted dataset?
image

I also left some comments for you to address. Can you please fix this and recheck if your summary component changes work as expected and if your fixes didnt impact the stepper component?

libs/damap/src/lib/services/summary.service.ts
```diff
@@ -232,24 +232,30 @@ export class SummaryService {
...
+    storageLevel.completeness = Math.max(
+      (storedDatasetsCount / datasets.length) * 100 - 25,
+      25,
+    );
+    if (storageLevel.completeness < 75) {
+      storageLevel.status.push('dmp.steps.summary.storage.missingstorage');
+    }
+
+    if (dmp.externalStorageInfo) {
+      storageLevel.completeness += 25;
```
ValentinFutterer: There is still a bug here, since you only need the externalStorageInfo when you are actually using external storages. But it a still flags it as incomplete when you are only using internal storages.

PR #463, opened 2025-05-23
feat: improve info card visibility

### LINKED ISSUE #461: Global: Info text is not visible when working on lower steps
**Describe the change**
When working on lower steps, this infotext is not visible anymore, especially on smaller screens. It should maybe have a fixed position, so its always visible, but other ideas are also welcome. The user could theoretically scroll up everytime he/she switches step, but who does that :) Ideas should probably discussed with the Team and Livia beforehand.

Image

**Is this a regression?**
No.

**Expected behavior**
Info text should always be visible, without much work from the user.

**Environment**
Damap version 4.5.0

**Additional context**
The points in this issue come from internal feedback collected from the ZDFM at TU Wien.

lpandath:
## Description

This PR introduces a sticky info-card that remains visible at the top of the viewport, addressing the issue where the info-card was not accessible when working on lower steps, especially on smaller screens.

#### What does this PR do?

Sticky Position: the info-card is fixed at the top during scrolling, ensuring constant visibility.
Collapse/Expand Functionality: users can minimize or expand the info-card as needed.
Styling Enhancements: added subtle shadow, appropriate z-index, and a background color.

#### Code review focus

Persistent access to the info-card without the need to scroll up.
Usability on smaller screens by allowing users to control the info-card's visibility.

#### Idea

Screenshot 2025-05-23 at 16 40 34

### Checks

- [ ] Summary updated
- [ ] Version view updated
- [ ] Documentation added
- [ ] Tests added
- [ ] E2e tests created

closes GH-461

### DIFF
--- a/libs/damap/src/lib/components/dmp/dmp.component.css
+++ b/libs/damap/src/lib/components/dmp/dmp.component.css
@@ -5,6 +5,7 @@
   gap: 20px;
   margin: 20px;
   padding-bottom: 80px;
+  padding-top: 10px;
 }
 
 .steps-container {
--- a/libs/damap/src/lib/widgets/info-card/info-card.component.css
+++ b/libs/damap/src/lib/widgets/info-card/info-card.component.css
@@ -3,13 +3,26 @@
   border: 1px solid #ccc;
   border-radius: 8px;
   margin: 0 20px;
+  position: sticky;
+  top: 0;
+  z-index: 1000;
+  background-color: var(--primary-lightest);
+  margin-bottom: 20px;
+  box-shadow: 0 2px 5px rgba(0, 0, 0, 0.1);
+}
+
+.card-controls {
+  display: flex;
+  justify-content: flex-end;
+  align-items: center;
+  position: absolute;
+  top: 0;
+  right: 0;
+  z-index: 2;
 }
 
 .close-button {
-  position: absolute;
-  top: 10px;
-  right: 10px;
-  z-index: 1;
+  position: relative;
   background: transparent;
   border: none;
   color: var(--neutral-color-medium-light);
@@ -70,6 +83,29 @@
   font-weight: 300;
   color: var(--user-navbar-color);
   padding: 16px 16px;
+}
+
+.info-box.collapsed {
+  padding: 8px;
+  cursor: pointer;
+}
+
+.info-box.collapsed .flex-row-container {
+  justify-content: center;
+}
+
+.info-box.collapsed .info-box-container {
+  display: none;
+}
+
+.expand-collapse-btn {
+  background: transparent;
+  border: none;
+  color: var(--primary-color);
+  cursor: pointer;
+  display: flex;
+  align-items: center;
+  margin-left: auto;
 }
 
 .mat-content {
--- a/libs/damap/src/lib/widgets/info-card/info-card.component.html
+++ b/libs/damap/src/lib/widgets/info-card/info-card.component.html
@@ -1,12 +1,20 @@
 <mat-card
   appearance="outlined"
   *ngIf="isIntroShow"
-  class="info-box mat-icon-shadow custom-card">
-  <button (click)="closeIntro()" mat-icon-button class="close-button">
-    <mat-icon class="material-icon-primary close-icon" matSuffix>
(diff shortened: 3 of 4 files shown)


ValentinFutterer:
First off, the collapse/expand does not work at all. Clicking the card does nothing and even if you fix that, the uncollapse button is also not functional. Please take the time to click through your code before you push soemthing and create a PR.

I fixed this part of the code and let Livia test it, since she has a background in UX/UI.
The feedback we got was that the collapsing is good and it should be the default state, after the info-bar leaves its original place due to scrolling. The autocollapse is necessary, since the vertical space gets already constricted from below with the action bar and now also from the top with the info bar. Uncollapsing should be done with hovering over the info bar.

image
The collapsed state should also be more sleek, to save vertical space. Livia suggested to make the step icon have the same size as the icons in the stepper (but keep it like is when uncollapsed). The show details button should be removed and the remove button can be replaced with our standard info icon, to show that there is some info here.
image

Additionally, the sticky info bar should be removed for small screens, since the loss of working area is too great here. A good cutoff-point would be 640px.

- [ ] Autocollapse after leaving "standard" position when scrolling
- [ ] Uncollapse when hovering
- [ ] Make step icon smaller when collapsed
- [ ] Replace buttons with info icon when collapsed
- [ ] Make new features only work over 640px height

libs/damap/src/lib/components/dmp/dmp.component.css
```diff
@@ -5,6 +5,7 @@
   gap: 20px;
   margin: 20px;
   padding-bottom: 80px;
+  padding-top: 10px;
```
ValentinFutterer, line 8: Is there a reason for adding padding here? It makes almost no visual difference.

lpandath:
> First off, the collapse/expand does not work at all. Clicking the card does nothing and even if you fix that, the uncollapse button is also not functional. Please take the time to click through your code before you push soemthing and create a PR.
> 
> I fixed this part of the code and let Livia test it, since she has a background in UX/UI. The feedback we got was that the collapsing is good and it should be the default state, after the info-bar leaves its original place due to scrolling. The autocollapse is necessary, since the vertical space gets already constricted from below with the action bar and now also from the top with the info bar. Uncollapsing should be done with hovering over the info bar.
> image The collapsed state should also be more sleek, to save vertical space. Livia suggested to make the step icon have the same size as the icons in the stepper (but keep it like is when uncollapsed). The show details button should be removed and the remove button can be replaced with our standard info icon, to show that there is some info here. image
> 
> Additionally, the sticky info bar should be removed for small screens, since the loss of working area is too great here. A good cutoff-point would be 640px.
> 
>     * [ ]  Autocollapse after leaving "standard" position when scrolling
> 
>     * [ ]  Uncollapse when hovering
> 
>     * [ ]  Make step icon smaller when collapsed
> 
>     * [ ]  Replace buttons with info icon when collapsed
> 
>     * [ ]  Make new features only work over 640px height

@ValentinFutterer there is no collapse/expand in the first place. I wanted to discuss with Livia therefore i posted in the chat.

About the rest i will check if that makes sense for the accessibility and continue on the implementation. Please include Livia in the GitHub or in our next meeting. Thank you!

libs/damap/src/lib/widgets/info-card/info-card.component.css
```diff
@@ -3,13 +3,26 @@
   border: 1px solid #ccc;
   border-radius: 8px;
   margin: 0 20px;
+  position: sticky;
+  top: 0;
+  z-index: 1000;
+  background-color: var(--primary-lightest);
+  margin-bottom: 20px;
```
ValentinFutterer, line 10: I dont get see you added this, am I missing something? Is it there so that there is more space to scroll before the content ist covered by the info-bar?

ValentinFutterer:
> @ValentinFutterer there is no collapse/expand in the first place. I wanted to discuss with Livia therefore i posted in the chat.

But there is. A lot of code in the PR is about collapse expand. If you didnt want me to look at it, why did you push all the collapse code and also write it in the PR description? if you just want feedback or start a discussion please create a draft PR and explicitly state which parts are not ready yet.

PR #475, opened 2025-06-30
Access: Add option to give ownership rights

### LINKED ISSUE #474: Access: Add Option to grant ownership
**Changes**
Right now, ownership cannot change and canot be given to other contributors. Add support for giving and taking away rights for owners. Owners should also be able to take away owner rights from other owners.

**Additional notes**
There should be a low amount of refactoring done, since we will touch this part of the code eagain in the fututre during big reworks. The issue is connected to the backend issue https://github.com/damap-org/damap-backend/issues/390

**Acceptance Criteria**
UI allows for giving Owner rights, editor rights and for taking away rights form contributors. Only owners should be able to

ValentinFutterer:
## Description

Feature

#### What does this PR do?

- Added radio buttons for assigning contributors rights to the DMP - now its also possible to assign ownership rights. Now only owners are allowed to assign and take away rights (previously editors could also give editor rights).
- Since the acess page is accessible to everyone that can fetch the DMP (bad design, should be changed in the future) contributors with editor rights could still see the access page over manipulating the URL. I prevented this by adding an alternative view with an info bar.
- Owners cant revoke their own ownership (circumvents the frontend validation for "there always has to eb at least one owner" and is just cheaper to do)
- To remove rights, i added the `NO_RIGHTS` role to the frontend. This change is *NOT* reflected in the backend, since the backend right now treats no rights as no access entity exists. This absolutely needs to be refactored, but this PR will not do it. Without the no rights role, its hard to use the radio buttons effectively.
- There is no warning when selectin owner, so users migth not know what this entails.

#### Code review focus

Security aspects. The surrounding code was not perfect and i refrained from a lot of otherwise necessary refactoring to save on time- this will be tackled on a later date.

#### Dependencies

[](https://github.com/damap-org/damap-backend/pull/399)

closes GH-474

### DIFF
--- a/libs/damap/src/assets/i18n/templates/en.json
+++ b/libs/damap/src/assets/i18n/templates/en.json
@@ -453,8 +453,8 @@
         "storage": {
           "allinfo": "All information necessary provided.",
           "nonestored": "No datasets will be stored, as no produced or reused datasets have been specified.",
-          "missingstorage": "Some data is not stored.",
-          "missingexplanation": "Explanation for using external storage is missing."
+          "missingstorage": "Some datasets do not have storages assigned.",
+          "missingexplanation": " Explanation for using external storage is missing."
         },
         "licensing": {
           "complete": "All information necessary provided.",
--- a/libs/damap/src/lib/auth/auth.service.ts
+++ b/libs/damap/src/lib/auth/auth.service.ts
@@ -53,4 +53,9 @@ export class AuthService {
   logout() {
     this.oAuthService.logOut();
   }
+
+  getId(): string {
+    const claims = this.oAuthService.getIdentityClaims();
+    return claims['personID'];
+  }
 }
--- a/libs/damap/src/lib/components/access/access.component.html
+++ b/libs/damap/src/lib/components/access/access.component.html
@@ -1,38 +1,51 @@
 <div id="access-wrapper">
-  <div *ngIf="dmp$ | async as dmp">
-    <h2>{{ dmp.project?.title || "DMP ID: " + dmp.id }}</h2>
-  </div>
-  <div class="action-buttons">
-    <button mat-raised-button [routerLink]="'/plans'">
-      <mat-icon>chevron_left</mat-icon>
-      {{ "access.button.plans" | translate }}
-    </button>
-    <button
-      mat-raised-button
-      class="button-color-primary"
-      [routerLink]="'/dmp/' + id">
-      {{ "versions.view.button.goto.latest" | translate }}
-    </button>
-  </div>
-  <app-info-message
-    >{{ "access.list" | translate
-    }}{{ "access.info" | translate }}</app-info-message
-  >
-  <div class="access-list" *ngFor="let access of accesses$ | async">
-    <damap-person-card [person]="access">
-      <div *ngIf="access.access !== accessType.OWNER; else owner">
-        <mat-checkbox
-          [checked]="access.access === accessType.EDITOR"
-          (change)="editorToggle($event, access)">
-          {{ "access.editor" | translate }}
-        </mat-checkbox>
-      </div>
-      <ng-template #owner>
-        <div class="owner">
-          <mat-icon class="mat-icon-style material-icon-primary">star</mat-icon>
-          {{ "access.owner" | translate }}
+  <div *ngIf="isOwner$ | async; else noOwner">
+    <div *ngIf="dmp$ | async as dmp">
+      <h2>{{ dmp.project?.title || "DMP ID: " + dmp.id }}</h2>
+    </div>
+    <div class="action-buttons">
+      <button mat-raised-button [routerLink]="'/plans'">
+        <mat-icon>chevron_left</mat-icon>
+        {{ "access.button.plans" | translate }}
+      </button>
+      <button
+        mat-raised-button
+        class="button-color-primary"
+        [routerLink]="'/dmp/' + id">
+        {{ "versions.view.button.goto.latest" | translate }}
+      </button>
+    </div>
+    <app-info-message
+      >{{ "access.list" | translate
+      }}{{ "access.info" | translate }}</app-info-message
+    >
+    <div class="access-list" *ngFor="let access of accesses$ | async">
+      <damap-person-card [person]="access">
+        <div *ngIf="access.universityId !== this.userId; else user">
+          <mat-radio-group
+            class="radio-buttons"
+            [(ngModel)]="access.access"
+            (change)="toggleAccess($event, access)">
+            <mat-radio-button *ngFor="let role of roles" [value]="role">
+              <span class="role-text">{{ role }}</span>
+              <app-tooltip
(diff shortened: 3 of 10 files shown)


libs/damap/src/lib/components/access/access.component.html
```diff
@@ -1,38 +1,49 @@
...
-        </mat-checkbox>
-      </div>
-      <ng-template #owner>
-        <div class="owner">
-          <mat-icon class="mat-icon-style material-icon-primary">star</mat-icon>
-          {{ "access.owner" | translate }}
+  <div *ngIf="isOwner$ | async; else noOwner">
+    <div *ngIf="dmp$ | async as dmp">
+      <h2>{{ dmp.project?.title || "DMP ID: " + dmp.id }}</h2>
+    </div>
```
BlueTechie, line 5: Guarding sensitive content with `*ngIf` is not sufficient since it only hides the UI and can be bypassed (e.g. by making an API call with Postman). I think it would be best if the information only flows to the client if the server validates that the user has the rights to see it. In the case that the server denies access, the frontend can display a notification/error.

libs/damap/src/lib/components/access/access.component.ts
```diff
@@ -52,22 +64,42 @@ export class AccessComponent implements OnInit {
...
+          .some(access => access.access === 'OWNER'),
+      ),
+    );
   }
 
   private getAccess(id: number) {
-    this.accesses$ = this.backendService.getAccess(id);
+    this.accesses$ = this.backendService.getAccess(id).pipe(
+      map(accesses => {
+        return accesses.map(access => {
```
BlueTechie, line 84: We should be careful here since currently the API endpoint `getAccessForDmp(@PathParam("id") String id)` in `AccessRessource.java` calls the `AccessValidator.java` which contains the following code:
```java
List<Access> accessList = accessRepo.getAccessByDmp(dmp);
    Optional<Access> dmpAccess =
        accessList.stream()
            .filter(
                access ->
                    access.getUniversityId().equals(securityService.getUserId())
                        && (access.getRole().equals(EFunctionRole.EDITOR) // <- should be removed
                            || access.getRole().equals(EFunctionRole.OWNER)))
            .findAny();
```
In particular all users with write permissions, i.e. editors, are able to view the access for a certain DMP instead of only the owner(s). If the goal is to only allow admins and owners to receive this access information, it would be best to remove this line in the backend and make the appropriate changes in the frontend if access is refused, as noted in my previous comment.

ValentinFutterer: True that was an oversight that happened early on and unfortunately i continued building some stuff on top of it 😅 
I will remove this from the backend so only owners can view accesses.

libs/damap/src/lib/components/access/access.component.html
ValentinFutterer: I will remove the code and add change the validation in the backend, like described in the comment below.

ValentinFutterer: Correction, the code can stay as is, changing the backend was enough.

bitmotivator (approved):

libs/damap/src/assets/i18n/templates/en.json
```diff
@@ -452,8 +452,8 @@
         "storage": {
           "allinfo": "All information necessary provided.",
           "nonestored": "No datasets will be stored, as no produced or reused datasets have been specified.",
-          "missingstorage": "Some data is not stored.",
-          "missingexplanation": "Explanation for using external storage is missing."
+          "missingstorage": "Some datasets do not have storages assigned.",
+          "missingexplanation": " Explanation for using external storage is missing."
```
bitmotivator, line 457: Why is there a whitespace here?

libs/damap/src/lib/auth/auth.service.ts
```diff
@@ -53,4 +53,9 @@ export class AuthService {
   logout() {
     this.oAuthService.logOut();
   }
+
+  getId(): string {
+    const claims = this.oAuthService.getIdentityClaims();
+    return claims['personID'];
```
bitmotivator, line 59: This should be fetched from the backend as it is configurable. It is available in the config endpoint under the `authUser` key.

libs/damap/src/lib/components/access/access.component.ts
```diff
@@ -52,22 +66,42 @@ export class AccessComponent implements OnInit {
...
       this.router.navigate(['/']);
     }
+
+    this.userId = this.authService.getId();
+
+    this.isOwner$ = this.accesses$.pipe(
+      map(accesses =>
+        accesses
+          .filter(access => access.universityId === this.userId)
+          .some(access => access.access === 'OWNER'),
```
bitmotivator: Isn't there a constant or enum value for something like this? It seems suboptimal to use this as a string. Like `FUNCTION_ROLES.OWNER` used in the same PR below.

bitmotivator (dismissed):
A few nit-picks.

libs/damap/src/assets/i18n/templates/en.json
ValentinFutterer: The text is sometimes appended to other text and then you need the whitespace. The whitespace was there before and i removed it erroneously - thats why i readded it. In case its only this text, the whitespace is not rendered.

libs/damap/src/lib/auth/auth.service.ts
ValentinFutterer: I think you got something mixed up here, how would you be able to preconfigure a users person ID in the backend. `authUser` just configures the variable name for userIDs - if this is needed is another question. 
The answer is no fyi, the property is not used anywhere in the frontend 😄 
We can remove this in a seperate PR.

libs/damap/src/lib/components/access/access.component.ts
ValentinFutterer: Good point :)

libs/damap/src/lib/auth/auth.service.ts
bitmotivator: Please add at least a todo that this shouldn't be hard-coded because it affects the ability to use OIDC systems other than Keycloak, so this is a regression.

libs/damap/src/assets/i18n/templates/en.json
bitmotivator: :vomiting_face:

ValentinFutterer: Yes 😄

libs/damap/src/lib/auth/auth.service.ts
ValentinFutterer: I see what you mean now, it should be configurable 😅 
Yes it should but not in this PR :)

bitmotivator (approved):

lpandath:
@bitmotivator i see that the PR is arelady closed, but when i opend the issue i thought that validate role should be changed permissions on server-side?

PR #483, opened 2025-08-06
Step 2: improve UI

### LINKED ISSUE #455: Contributors: Small UI improvements
**Changes**
After adding a contributor via the search bar, the search query stays in the search bar. It should be cleared after adding.
Image

This icon confused testers. Sadly there is no great fit for what we wanted to show with the icon. It was suggested to use the neutral "Stat Minus 2" icon.
Image

Name and email are required fields, but they are missing stars.
Image

On a smaller screen, those two fields are not inline, even though there is enough space.
Image

There is a colon missing here.
Image

**Additional notes**
The points in this issue come from internal feedback collected from the ZDFM at TU Wien.

DanceaVlad:
## Description

Feature

#### What does this PR do?

1. In People involved in data management step -> Search contributor -> Find Contributors:

- type in **Search for person...**
- select contributor 
- NEW: text is removed from search field

2. In People involved in data management step -> Search contributor -> Project contact:

- change icon from **person_remove** to **stat_minus_2**

3. In People involved in data management step -> Input contributor manually -> Additional contributors: 

- show a star (*) next to First name, Last name and email to show they are mandatory

#### Code review focus

1. For addition nr. 1, I think it would make sense to also close the autocomplete pane after selecting a contributor, not just removing the text in the search field.

2. (Not done) In People involved in data management step -> Input contributor manually -> Additional contributors: 

- as the screen in becoming smaller, have the 2 fields change to vertical alignment later
- this is very janky. By using media queries it either works for when the side-menu is open or when it is closed, but you cannot make it work for both situations.
- you either need an observable for the side-menu and apply conditional css, or another solution:
CleanShot 2025-08-06 at 15 19 43@2x

- This works perfectly, independent of the side-menu state, as it uses the container size, not the screen size
- The problem with this is that @container media query is only supported for browser versions of the past 2-3 years.
- This is a diff that solves the problem, but from what I heard the side-menu is an overarching issue that may need an overhaul in the future. Tell me and I'll add this code to the PR.
[css.patch](https://github.com/user-attachments/files/21619734/css.patch)

closes GH-455

### DIFF
--- a/libs/damap/src/assets/i18n/templates/en.json
+++ b/libs/damap/src/assets/i18n/templates/en.json
@@ -40,7 +40,7 @@
         "intro": "In this step, you add people involved in the project and provide additional info about them, e.g., their role. In this example, the project leader’s info has already been added and marked as the contact person (note icon with person and envelope next to their name). You can add more contributors via the search field (ensure you select the appropriate database, e.g., ORCID), or manually via the “Add contributor manually button”. Please provide a minimum of one contact person as well as a data manager (it can be the same person).",
         "search": "Search for person...",
         "search.service": "Search service",
-        "general.search": "Find Contributors",
+        "general.search": "Find Contributors:",
         "general.recommendations": "Recommended contributors:",
         "general.addAll": "Add all contributors",
         "contributor": {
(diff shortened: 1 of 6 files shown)


ValentinFutterer:
> For addition nr. 1, I think it would make sense to also close the autocomplete pane after selecting a contributor, not just removing the text in the search field.

You are correct, please add it :) 

Regarding the alignment, we can keep it as is for now - this is a whole other can of worms we dont necessarily need to open know.

ValentinFutterer (approved):
Good work 👍

PR #487, opened 2025-08-25
Make project service nullsafe

ValentinFutterer:
## Description

Bugfix

#### What does this PR do?

When you are using the default project service and then switch to Pure, you get a null pointer. This PR fixes it.

This PR also fixes a small typo in the github workflow test.yml (thanks to Sotirios for spotting it immediately)


BlueTechie (approved):
The nullpointer exception has already been fixed in a separate PR. However, defensive programming is a good idea and the typo fix is valid 👍

PR #489, opened 2025-09-17
People: Fix wrong servicetype being used

ValentinFutterer:
## Description

Bugfix

#### What does this PR do?

Sends the correct queryvalue instead of the displaytext. This can cause bugs with the Pure integration and University service, since display is lowercase and query is all uppercase. To test it out move ORCID to first place in the person services. the frontend will now only use  the ORCID service (there is a fallback in the backend, which always chooses the first one when the service type is not found)


lpandath (approved):
LGTM to me!

Issue #411, opened 2025-01-09
Clarify contact person definition and role assignment in find person and search service


lpandath:
**Changes**
![Screenshot 2025-01-09 at 16 55 36](https://github.com/user-attachments/assets/690a2848-1a9a-47b3-9367-24aa1ae546f1)

1. Clearly show that users need to define a contact person by using the contact symbol on the right.
2. Add instructions or a tooltip to help users understand how to assign roles to the contact person.
3. Make the process of setting up a contact person and assigning roles easier and more intuitive, so users can find the settings quickly.

**Additional notes**
The current UI is unclear about where to define the contact person and assign roles.

**Assumptions**
Consider tooltips or instructions are added to help users understand the process.

**Acceptance Criteria**
Users can clearly see where to define a contact person and know how to assign roles.

**Risks? (optional)**
Potential risks and how to best mitigate them.

ValentinFutterer:
Contact person UI should be made more intuitive by [this](https://github.com/tuwien-csd/damap-frontend/pull/394)
I dont get what is meant by 2. The contact person is a normal contributor, so roles can be assigned like normal.

lpandath:
@ValentinFutterer was still not that intuitive, but we can of course get more feedback from other users. 
Another point: the role form field could contain a tooltip or similar to inform you about the role description.

lpandath:
![Image](https://github.com/user-attachments/assets/b64dd44a-3557-4383-8ff0-137bf39f0e04)

@ValentinFutterer @andresTabiTuwien 
Alternative UI Enhancements:

_first_:
When the user clicks into the search field (without typing anything), the recommended contributors list is displayed below the search field. This list remains visible until the user starts typing.
_Typing Behavior_: as soon as the user starts typing (on keypress), the recommended contributors disappear, and the actual search results are shown.
_Search Completion Behavior:_ If the user clears the search input (deletes all characters), the recommended contributors list reappears automatically.

_second_:
Collapsible recommended contributors section. The recommended contributors should not be permanently visible.
They are inside a collapsible container (similar to an accordion):
Initially open by default.
If the user starts typing or selects a contributor, the container automatically collapses.
After collapsing, it should not reopen automatically unless triggered manually.
The user can manually reopen it by clicking a “Show Recommendations” button.

_not that optimal:_
Instead of stacking sections, use a two-column layout:
Left Side: search & recommended contributors
Right Side: selected contributors (updated in real-time)

[...]

PR #491, opened 2025-10-01
eduID integration, Authentication overhaul

GeoffreyKarnbach:
Frontend PR for Backend:

https://github.com/damap-org/damap-backend/issues/411

Adds new login flow, HTTP Authentication Interceptor, changes to the Guard system, Auth Service, Replacing OIDC by the custom JWT logic, add new service to add persons to a DMP

### DIFF
--- a/apps/damap-frontend/src/app/services/config.service.ts
+++ b/apps/damap-frontend/src/app/services/config.service.ts
@@ -2,8 +2,8 @@ import { AuthConfig, OAuthService } from 'angular-oauth2-oidc';
 import { BehaviorSubject, Observable, lastValueFrom } from 'rxjs';
 import { Injectable, isDevMode } from '@angular/core';
 
-import { Config } from '@damap/core';
 import { FeedbackService } from '@damap/core';
+import { BackendService, Config } from '@damap/core';
 import { HttpClient } from '@angular/common/http';
 import { Router } from '@angular/router';
 import { environment } from '../../environments/environment';
@@ -21,6 +21,7 @@ export class ConfigService {
     private http: HttpClient,
     private oauthService: OAuthService,
     private router: Router,
+    private backendService: BackendService,
     private feedbackService: FeedbackService,
   ) {}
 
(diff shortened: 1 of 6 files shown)


apps/damap-frontend/src/app/services/config.service.ts
```diff
@@ -35,34 +36,103 @@ export class ConfigService {
...
-                if (url !== '') {
-                  return this.router.navigateByUrl(url);
-                }
+                this.backendService
+                  .getAuthJWT(this.oauthService.getAccessToken())
+                  .subscribe(
+                    response => {
+                      localStorage.setItem(
+                        'damap_session_token',
+                        response.token,
```
BlueTechie, line 75: Session tokens should not be stored in localStorage ([https://auth0.com/docs/secure/security-guidance/data-security/token-storage#don-t-store-tokens-in-local-storage](https://auth0.com/docs/secure/security-guidance/data-security/token-storage#don-t-store-tokens-in-local-storage) has more info). Instead we should use a secure session cookie.

apps/damap-frontend/src/app/services/config.service.ts
```diff
@@ -35,34 +36,103 @@ export class ConfigService {
...
-            // sessionChecksEnabled: true,
-          };
-          this.oauthService.configure(authConfig);
-          this.oauthService.setupAutomaticSilentRefresh();
-          return this.oauthService
-            .loadDiscoveryDocumentAndTryLogin()
-            .then(() => {
+
+          if (config.env === 'PROD') {
+            const manualAuthConfig: AuthConfig = {
```
BlueTechie, line 49: See my comment in the backend PR concerning `env`.

libs/damap/src/lib/auth/auth.service.ts
```diff
@@ -1,61 +1,100 @@
...
+    return claims['sub'];
+  }
+
+  getClaimsFromJWT(): any {
+    const jwtToken = localStorage.getItem('damap_session_token');
+    if (!jwtToken) {
+      return null;
+    }
+    const claims = JSON.parse(atob(jwtToken.split('.')[1]));
+    return claims;
```
BlueTechie, line 98: There is a signature check of the JWT missing. Without it, anyone could forge a token with any payload they want and claim that they are an admin, among other possibilities.

libs/damap/src/lib/auth/auth.service.ts
```diff
@@ -1,61 +1,100 @@
...
+    const claims = this.getClaimsFromJWT();
+    if (!claims) {
+      return '';
+    }
+    return claims['sub'];
+  }
+
+  getClaimsFromJWT(): any {
+    const jwtToken = localStorage.getItem('damap_session_token');
+    if (!jwtToken) {
```
BlueTechie, line 94: See my other comment about using `localStorage` to store the session token.

libs/damap/src/lib/interceptors/token.interceptor.ts
```diff
@@ -0,0 +1,63 @@
...
+
+  intercept(
+    request: HttpRequest<any>,
+    next: HttpHandler,
+  ): Observable<HttpEvent<any>> {
+    const token = localStorage.getItem('damap_session_token');
+
+    if (!token) {
+      return next.handle(request);
+    }
```
BlueTechie, line 30: Currently, the token is sent to any outgoing HTTP request. We should modify this to
`if (!token || !request.url.startsWith(environment.backendurl))` to only do so when communicating with the backend (consider for instance the redirects to other non-damap URLs, such as to find out more about licenses).

libs/damap/src/lib/interceptors/token.interceptor.ts
```diff
@@ -0,0 +1,63 @@
...
+    private backendService: BackendService,
+  ) {}
+
+  private isRefreshing = false;
+
+  intercept(
+    request: HttpRequest<any>,
+    next: HttpHandler,
+  ): Observable<HttpEvent<any>> {
+    const token = localStorage.getItem('damap_session_token');
```
BlueTechie, line 26: See my other comment on localStorage.

libs/damap/src/lib/services/backend.service.ts
```diff
@@ -461,6 +461,18 @@ export class BackendService {
     return this.http.delete<void>(`${this.backendUrl}admin/banner`);
   }
 
+  getAuthJWT(accessToken: string): Observable<{ token: string }> {
+    return this.http.get<{ token: string }>(
+      `${this.backendUrl}auth/jwt?access_token=${accessToken}`,
+    );
```
BlueTechie, line 472: See my other comment in the backend PR: We should not be sending `accessToken` as part of the URL. Instead, we should use POST.

BlueTechie (changes requested):
I have found some vulnerabilities that are linked to the architecture choices of the backend, and some that are frontend specific. As in the backend PR, fixing them should not be too time consuming.

libs/damap/src/lib/auth/auth.service.ts
GeoffreyKarnbach: Worst case the user can see the admin UI, but cannot actually perform any actions, as the backend itself verifies the signature of the JWT independently.

libs/damap/src/lib/interceptors/token.interceptor.ts
GeoffreyKarnbach: Great catch!

libs/damap/src/lib/services/backend.service.ts
GeoffreyKarnbach: Fixed

apps/damap-frontend/src/app/services/config.service.ts
GeoffreyKarnbach: In the previous implementation of DAMAP relying on KeyCloak we used the same mechanism. Changing this up now, would require major changes in the processing of requests, especially in the backend. Is there any less invasive alternative?

GeoffreyKarnbach: We discussed it, as we use localstorage only for short lived tokens of under 3 hours and localstorage might only be an issue in case of XSS (which we anyways need to strongly prevent), LocalStorage is suitable to be kept.

BlueTechie (approved):

libs/damap/src/lib/auth/auth.service.ts
```diff
@@ -1,61 +1,100 @@
...
+      return false;
+    }
+    return claims['groups']?.includes('Damap Admin');
   }
 
   logout() {
-    this.oAuthService.logOut();
+    localStorage.removeItem('damap_session_token');
+    // Navigate to start page, remove anything after first /
+    this.router.navigateByUrl('/');
```
ValentinFutterer: This doesnt log you out, it just moves you to the landing page

ValentinFutterer, line 81: For Keycloak: This doesnt log you out, it just moves you to the landing page

apps/damap-frontend/src/app/services/config.service.ts
```diff
@@ -43,34 +44,103 @@ export class ConfigService {
...
-                }
+                this.backendService
+                  .getAuthJWT(this.oauthService.getAccessToken())
+                  .subscribe(
+                    response => {
+                      localStorage.setItem(
+                        'damap_session_token',
+                        response.token,
+                      );
+                      this.oauthService.logOut(true);
```
ValentinFutterer, line 77: Can you explain why this is needed, I cant figure it out 😅

apps/damap-frontend/src/app/services/config.service.ts
```diff
@@ -43,34 +44,103 @@ export class ConfigService {
...
-            showDebugInformation: isDevMode(),
-            // sessionChecksEnabled: true,
-          };
-          this.oauthService.configure(authConfig);
-          this.oauthService.setupAutomaticSilentRefresh();
-          return this.oauthService
-            .loadDiscoveryDocumentAndTryLogin()
-            .then(() => {
+
+          if (config.env === 'PROD') {
```
ValentinFutterer, line 48: Is there a way to make this code shorter, it shares a lot of code with the Dev is branch
Although this could be tricky, i realise, its just a thought

ValentinFutterer (commented):
When you log in, you get redirected to the landing page, before you then go to the homepage.

apps/damap-frontend/src/app/services/config.service.ts
lpandath, line 48: Here the SPA sends a dummySecret with useHttpBasicAuth. A browser client can not keep secrets. For eduID/public clients this should be PKCE-only. Can we drop the secret/basic auth here and confirm backend #411 works with pure PKCE?

libs/damap/src/lib/interceptors/token.interceptor.ts
```diff
@@ -0,0 +1,64 @@
...
+import { environment } from '../../../../../apps/damap-frontend/src/environments/environment';
+
+@Injectable()
+export class TokenInterceptor implements HttpInterceptor {
+  constructor(
+    private authService: AuthService,
+    private backendService: BackendService,
+  ) {}
+
+  private isRefreshing = false;
```
lpandath, line 20: while isRefreshing is true, concurrent requests still go out with the expired token into 401s/logouts

ValentinFutterer:
Redundant due to https://github.com/damap-org/damap-frontend/pull/511

PR #492, opened 2025-10-13
Textarea: Fixed visibility of questions

### LINKED ISSUE #490: Question in text field no longer visible when entering text
**Describe the bug**
see under "Documentation and data quality",
Question in text field no longer visible when entering text,
Concerns question 2. "Please indicate which metadata...." and question 3. "How will you provide documentaion..."
also concerns section Reuse of data/ question 1+2;
maybe also other sections/text fields are concerned - did not check in detail;

Image

Image

**Is this a regression?**
Yes/No.

**Steps to Reproduce**

**Expected behavior**
question should always be visible when entering text in the text field. 

**Screenshots**

Image

Image

**Environment**
 - Browser Mozilla Firefox
 - Damap version 4.6.0

**Additional context**

BlueTechie:
## Description
Bugfix

#### What does this PR do?
Before changes, questions that required user input in the form of a textarea were cut off on smaller screens or when the user zoomed in. This PR places the question as separate text above the text area and shortens its label.

**Before** (Question should read: _Please indicate which metadata will be provided to help others identify, discover and reuse the data_):
image

**After changes**:
image

UPDATE List of additional changes:
- Removed some unused `css` classes and did some small refactoring in `licenses.component.css`, `reuse.component.css`, `specify-data-component.css`.
- Added the long label for textarea in the textarea wrapper component, which can be passed along as an input. This avoids having custom css code for each dmp step for the textareas.
- Moved the `dataGeneration` translation to `dmp.steps.data.specify.question` instead of `dmp.steps.documentation.question` (with a short and long version).
- Split every textarea prompt into a long and short version (except for some trivial ones).

closes GH-490


lpandath (approved):
looks just great! LGTM

ValentinFutterer (commented):
While this solution solves the problem, its really cumbersome and adds unnecessary bloat. I would suggest to move the `<mat-hint>` part into the `textarea-wrapper` component. There you can handle styling in one place and you save yourself all the unnecessary css classes.
Or you could make your life even easier and put the `<mat-hint>` inside the `<mat-form-field` and use the components styling. This way you dont need to manage anything (you already set the subscriptsizing to dynamic). Only issue here is, that the text is below the text field. You could check with Christiane/Livia if its ok, but I think it should be, since the shorter label text is still there.

Or, if you want to be fancy, show the label only if the textfield is empty and not in focus. If the textfield has a value or is in focus, use the `<mat-hint>`. This way, we have the best of both worlds - disclaimer I did not test if it would work.

ValentinFutterer:
Ah and @BlueTechie please link up with @DanceaVlad and @lpandath, since you are introducing new entries into the translation files.

libs/damap/src/lib/components/dmp/doc-data-quality/doc-data-quality.component.html
```diff
@@ -10,28 +10,36 @@
...
 
+        <mat-hint>{{
+          "dmp.steps.documentation.question.long.dataQuality" | translate
+        }}</mat-hint>
         <mat-form-field class="full-width" [formGroup]="docDataStep">
           <mat-label
-            >{{ "dmp.steps.documentation.question.dataQuality" | translate }}
+            >{{
+              "dmp.steps.documentation.question.short.dataQuality" | translate
+            }}
```
ValentinFutterer: Good catch on the overflow here. Can you maybe also adjust the space between the hint and the input field, so that its consistent witht he textareas?

Image

libs/damap/src/lib/components/dmp/licenses/licenses.component.html
```diff
@@ -107,11 +108,12 @@ <h4>{{ dataset.value.title }}</h4>
...
   </ng-container>
 </div>
-<div [style.margin.px]="'5'">
+<div class="summary-container">
   <div [formGroup]="dmpForm">
     <ng-container *ngIf="restricted?.length > 0">
       <app-textarea-wrapper
-        [label]="'dmp.steps.licensing.question.restrictedAccessInfo'"
+        [label]="'dmp.steps.licensing.question.short.restrictedAccessInfo'"
+        [longLabel]="'dmp.steps.licensing.question.long.restrictedAccessInfo'"
```
ValentinFutterer, line 116: This happens with a long dataset name:

Image

I would suggest to add the dataset names to the long info on top and not to the short label.

libs/damap/src/lib/components/dmp/licenses/licenses.component.html
```diff
@@ -120,7 +122,8 @@ <h4>{{ dataset.value.title }}</h4>
 
     <ng-container *ngIf="closed?.length > 0">
       <app-textarea-wrapper
-        [label]="'dmp.steps.licensing.question.closedAccessInfo'"
+        [label]="'dmp.steps.licensing.question.short.closedAccessInfo'"
+        [longLabel]="'dmp.steps.licensing.question.long.closedAccessInfo'"
```
ValentinFutterer, line 126: Same as above

libs/damap/src/lib/components/dmp/reuse/reuse.component.html
ValentinFutterer, line 31: This part should also use the shared textarea component and also needs the long and short version + it has the same problems as the closed and restricted textareas with long dataset names.

Image

ValentinFutterer (approved):
👍

PR #493, opened 2025-10-15
Dataset-sizes: Adapt ranges to backend

ValentinFutterer:
## Description

Bugfix

#### What does this PR do?

Backend had a bug introduced , where it translated the size ranges from the frontend wrong and bumped them up one step, due to a mismatch in ranges. (Check out DatasetSizeRange in the backend project) Since the frontend side would be easier to adapt and fix, I did it here.

This has now the effect that old DMPs have their size ranges bumped up one step 8this already happended on export, so it just gets clearer). Fixing it in the database would be too much of a hassle. It also introduced this sideeffect on old DMPs, where it shows nothing as preselected in the edit dataset window:

image>


lpandath (approved):
LGMT to me.

Tested with new DMPs: sizes save/display/edit - works
Tested with old DMPs: sizes display correctly

PR #505, opened 2025-12-10
Angular: Patch vulnerabilities with version upgrades of dependencies

BlueTechie:
## Description

Security patch
#### What does this PR do?
Update Angular dependencies to patch several high- and medium-severity vulnerabilities in the frontend.

If you find yourself in dependency hell while testing the changes from this PR, this may help:
```
del [path_to_package-lock.json]
rmdir /s /q node_modules  
npm cache clean --force        
npm install         
```


package.json
```diff
@@ -26,19 +26,19 @@
...
-    "@angular/forms": "19.2.10",
-    "@angular/material": "19.2.15",
-    "@angular/material-experimental": "19.2.15",
-    "@angular/material-moment-adapter": "19.2.15",
-    "@angular/platform-browser": "19.2.10",
-    "@angular/platform-browser-dynamic": "19.2.10",
-    "@angular/router": "19.2.10",
+    "@angular/animations": "19.2.17",
+    "@angular/cdk": "19.2.17",
+    "@angular/cli": "19.2.17",
```
lpandath: @angular-devkit/build-angular is bumped to ^19.2.19 while @angular-devkit/core/schematics stay at 19.2.9 and @angular/language-service at 19.2.10, with @angular/cli at 19.2.17. 

Should these be aligned to the same patch level to avoid build/test?

lpandath (commented):
align Angular devkit/CLI/language-service versions: example all to 19.2.17 or the chosen patch, and consider pinning @angular-devkit/build-angular instead of using ^.

package.json
lpandath: Also package.json (and lockfile) now report 4.6.0, we should update this since now 4.7.0

BlueTechie:
Thanks for your feedback! I have now updated all current vulnerable dependencies after some struggles with dependency hell. There is more work needed in the future, since there are deprecation warnings concerning `karma` and `eslint`. However, this will require some version upgrades with breaking changes and is outside of the scope of this vulnerability patch PR.

Note that `build-angular` needs to be on version `19.2.19` instead of `19.2.17` like the other `@angular-devkit/`  and `@angular/`  dependencies, since otherwise there would be a transitive dependency of a vulnerable `vite` version. Bumping up all versions from `19.2.17` to `19.2.19` is also not possible, as some angular dependencies do not have such a newer version.

GeoffreyKarnbach:
Works great! Good job patching the dependencies against the vulns.

package.json
ValentinFutterer, line 3: Instead of updating it manually, you could just rebase.

ValentinFutterer (commented):
Seems to run well, thanks for going through the dependency related pain 👍 

When i run npm install, my `package-lock.json` gets updated at line 51 under devDependencies:
`"@angular-devkit/build-angular": "^19.2.19",` becomes
`"@angular-devkit/build-angular": "19.2.19",`

Can you take care of this?

ValentinFutterer (approved):
Nice, lets merge

PR #514, opened 2026-01-19
Gk/eduid integration

GeoffreyKarnbach:
## Description
This PR adapts the frontend to the new backend identity architecture. It updates the Access Management tab to decouple access rights from project contributors, introducing a user search for granting permissions to registered users.

#### What does this PR do?
* **Access Management:**
    * Refactors `AccessComponent` to search for users via the new `GET /access/user-search` endpoint using an autocomplete input.
    * Decouples the `Access` interface from `Contributor`. Access objects now carry their own `firstName`, `lastName`, and `mbox`.
* **Data Model Refactoring:**
    * Renames `universityId` to `identifier`.
    * Renames `access` (enum) to `role`.
    * Adds `UserDo` interface for search results.
* **Component Updates:**
    * **PersonCard:** Now accepts `Contributor | Access` union type. Added type guards (getters) to safely access fields specific to contributors (e.g., `personId`), preventing runtime errors.
    * **SearchField:** Extended to accept an optional `[matAutocomplete]` input, allowing it to be reused for the user search dropdown.
* **Config:** Syncs `ConfigService` properties with the latest backend authentication changes.

#### Breaking changes
* **Interface Changes:** The `Access` interface no longer extends `Contributor`. Fields `access` and `universityId` have been renamed to `role` and `identifier` respectively.
* **Workflow:** Users must now be explicitly searched and invited via the Access tab; they are no longer implicitly listed based on project contributor metadata.

#### Dependencies
Requires Backend PR: https://github.com/damap-org/damap-backend/pull/454
- [x] Linting passed
- [x] Tests passed
- [x] Documentation added
- [ ] E2e tests created

closes GH-```


libs/damap/src/lib/domain/config.ts
```diff
@@ -4,14 +4,15 @@ export interface Config {
...
+  readonly userRolesClaim: string;
   readonly userIdClaim: string;
   readonly nameClaim: string;
   readonly givenNameClaim: string;
   readonly familyNameClaim: string;
   readonly emailClaim: string;
-  readonly adminRoleName: string;
   readonly responseType: string;
+  readonly userRolesClaimPath: string;
+  readonly adminRoleName: string;
```
ValentinFutterer, line 16: I think that was by mistake.

apps/damap-frontend/src/app/services/config.service.ts
```diff
@@ -134,6 +134,10 @@ export class ConfigService {
     return this.config?.emailClaim || null;
   }
 
+  public getUserRolesClaimPath(): string {
+    return this.config?.userRolesClaimPath || null;
+  }
+
```
ValentinFutterer, line 143: I think that was by mistake.

libs/damap/src/lib/domain/access.ts
```diff
@@ -1,9 +1,19 @@
...
   start?: Date;
   until?: Date;
+  readonly universityId: string;
+  readonly firstName: string;
+  readonly lastName: string;
+  readonly mbox: string;
+}
+
+export interface AccessUserDO {
+  subjectId: string;
```
ValentinFutterer: Same as in backend, please rename

libs/damap/src/lib/domain/access.ts
```diff
@@ -1,9 +1,19 @@
 import { FunctionRole } from './enum/function-role.enum';
-import { Contributor } from './contributor';
 
-export interface Access extends Contributor {
+export interface Access {
+  id: number;
   dmpId: number;
```
ValentinFutterer, line 5: Should also be readonly

libs/damap/src/lib/widgets/person-card/person-card.component.ts
```diff
@@ -21,7 +22,11 @@ import { TranslateModule } from '@ngx-translate/core';
...
 })
 export class PersonCardComponent {
-  @Input() person: Contributor;
+  @Input() person: Contributor | Access;
 
   readonly identifierType: any = IdentifierType;
+
+  get contributor(): Contributor {
+    return this.person as Contributor;
+  }
```
ValentinFutterer, line 47: I get why you did this - but its incredibly dangerous. If someone changes a field name ina ccess, e.g. renames mbox so its consistently named email across the application, this implementation breaks silently.

libs/damap/src/assets/i18n/access/en.json
```diff
@@ -6,7 +6,7 @@
     "no_rights": "Default setting. The user will not be able to see or interact with the DMP.",
     "editor": "Editor rights grant the user permissions to see the DMP and make changes to it. There are no restrictions on the kind of changes this user can make, bar deleting the DMP. The user can not however, assign viewing or editing permissions.",
     "owner": "Owners can view, write, delete and assign permissions for DMPs. They are also capable of creating new owners - handle with care, new owners can remove old ones.",
-    "list": "Only contributors in your universities database can be assigned editing rights for a DMP. ",
+    "list": "Only users that have already logged into DAMAP can be assigned editing rights for a DMP. ",
```
ValentinFutterer: I would change this wording, since institutional tools are not named DAMAP.

```suggestion
    "list": "Only users that have already logged in once can be assigned editing rights for a DMP. ",
```

libs/damap/src/lib/components/access/access.component.html
```diff
@@ -19,6 +19,23 @@ <h2>{{ dmp.project?.title || "DMP ID: " + dmp.id }}</h2>
...
+      <mat-icon matSuffix>search</mat-icon>
+      <mat-autocomplete
+        #auto="matAutocomplete"
+        [displayWith]="displayUser"
+        (optionSelected)="addNewAccess($event)">
+        <mat-option *ngFor="let user of searchResult$ | async" [value]="user">
+          {{ user.nickname }} <small>({{ user.email }})</small>
+        </mat-option>
+      </mat-autocomplete>
+    </mat-form-field>
```
ValentinFutterer: We actually have a `search-field.component` you could/should use here. You only need to expand it to support the `[matAutocomplete]`.
The mat-autocomplete tag can be used outside of the mat-form-field, if I am not mistaken.

I would also suggest to reuse the result styling from the person search, minus the person identifier, to keep this consistent across the application

libs/damap/src/lib/components/access/access.component.ts
```diff
@@ -37,20 +51,33 @@ import { TooltipModule } from '../../widgets/tooltip/tooltip.module';
...
+  accesses$ = this.accessesSubject.asObservable();
   isOwner$: Observable<boolean>;
   dmp$: Observable<Dmp>;
   id: number;
   protected readonly FunctionRole = FunctionRole;
   roles = Object.values(FunctionRole);
   userId: string;
 
+  searchControl = new FormControl('');
+  searchResult$: Observable<AccessUserDO[]>;
```
ValentinFutterer: Minor: The code deals with both accesses and users, it would improve readability a little to rename this to userSearchControl.

libs/damap/src/lib/components/access/access.component.ts
```diff
@@ -37,20 +51,33 @@ import { TooltipModule } from '../../widgets/tooltip/tooltip.module';
...
+
+  currentAccessList: Access[] = [];
```
ValentinFutterer: I am not quite sure why this is needed - it just adds extra complexity, doesnt it? `accessesSubject` is a behaviour subject and as such stores the latest emitted value, which is essentially what currentAccessList does right now.

libs/damap/src/lib/components/access/access.component.ts
```diff
@@ -76,22 +103,46 @@ export class AccessComponent implements OnInit {
...
-          if (access.access === null || access.access === undefined) {
-            access.access = FunctionRole.NO_RIGHTS;
-          }
-          return access;
-        });
+    this.searchResult$ = this.searchControl.valueChanges.pipe(
+      debounceTime(300),
+      distinctUntilChanged(),
+      switchMap((term: string) => {
+        if (term && term.length >= 3) {
```
ValentinFutterer: I would maybe lower this to >= 2 or remove it completely. When i first tried the feature out I thought it was broken. Maybe talk to Livia and ask if >= 2 is ok or if it needs some kind of indication, if you keep it.

libs/damap/src/lib/components/access/access.component.ts
```diff
@@ -76,22 +103,46 @@ export class AccessComponent implements OnInit {
...
+            );
+        }
+        return of([]);
       }),
     );
   }
 
+  protected displayUser(user: AccessUserDO): string {
+    return user ? user.nickname : '';
+  }
```
ValentinFutterer: This and `[displayWith]="displayUser"` in the html is not needed - You already have a custom option on how to display the results.

libs/damap/src/lib/components/access/access.component.ts
```diff
@@ -104,4 +155,30 @@ export class AccessComponent implements OnInit {
...
+    const user: AccessUserDO = event.option.value;
+    this.searchControl.setValue('');
+
+    let firstName = user.nickname;
+    let lastName = '';
+    if (user.nickname && user.nickname.includes(' ')) {
+      const parts = user.nickname.trim().split(/\s+(.*)/);
+      firstName = parts[0];
+      lastName = parts[1] || '';
+    }
```
ValentinFutterer: Like I already suggested in the backend, just store first and lastname in the user to prevent this splitting by hand.

ValentinFutterer (commented):
Looks pretty good, most comments are pretty minor or are only about code cleanup / consistency in the project.

libs/damap/src/lib/domain/config.ts
GeoffreyKarnbach: I removed the userRolesClaim as it was not used, i assume you mean that?

libs/damap/src/lib/components/access/access.component.ts
GeoffreyKarnbach: Removed

libs/damap/src/lib/domain/access.ts
GeoffreyKarnbach: Cannot be made readonly as it is being edited in access.component.ts

libs/damap/src/lib/widgets/person-card/person-card.component.ts
GeoffreyKarnbach: I added a failsafe guard, that returns undefined if the values do not exist and a comment at the top of the file explaning the rationale.

apps/damap-frontend/src/app/services/config.service.ts
BlueTechie: Not sure if this has been fixed: the only change since this comment is the removal of the `getUserRolesClaim()` function (not `getUserRolesClaimPath()`).

libs/damap/src/lib/components/access/access.component.ts
```diff
@@ -72,36 +98,78 @@ export class AccessComponent implements OnInit {
...
         .createAccess(access)
         .subscribe({ next: response => (access.id = response.id) });
     }
   }
+
+  addNewAccess(event: MatAutocompleteSelectedEvent): void {
+    const user: UserDo = event.option.value;
+    this.userSearchControl.setValue('');
+
+    console.log('Selected user from autocomplete:', user);
```
BlueTechie: I noticed this leftover from development:

ValentinFutterer (approved):
Ready to merge 👍

PR #516, opened 2026-02-10
Security: Upgrade 3 vulnerable dependencies

BlueTechie:
## Description
Security patch

#### What does this PR do?
Update 4 vulnerable npm dependencies that were introduced transitively:
- `tar` (override)
- `webpack` (override)
- `diff` (override)
- `lodash` (by regenerating `package-lock.json`)

Unfortunately, using the npm `override` feature is the only way to update the 3 above dependencies in a quick manner: The only other way would involve upgrading our angular version from 19 to 21, which introduces breaking changes.
#### Dependencies

closes GH-

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -360,108 +360,6 @@
         }
       }
     },
-    "node_modules/@angular-devkit/build-angular/node_modules/@angular/build": {
-      "version": "19.2.19",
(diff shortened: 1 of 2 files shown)


ValentinFutterer (approved):
Good catches, we will take care of removing the override when we do the next angular upgrade 👍

PR #517, opened 2026-02-11
Patch GitHub security warnings

BlueTechie:
## Description
Small security patches

#### What does this PR do?
Fix the following CodeQL and SonarCloud security warnings:
- Replace `ADD` in the Dockerfile with `COPY`. `ADD` is essentially `COPY` but with 2 extras: fetch-from-URL functionality and automatic TAR file extraction. These 2 extras are not needed here since we only copy a configuration file from the build context.
- Specified (and narrowed) `GITHUB_TOKEN` permissions for the continuous integration workflow. The job name was also changed from `build` to `continuous-integration` to be less confusing.

closes GH-

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -18,7 +18,8 @@ on:
   workflow_dispatch:
 
 jobs:
-  continuous-integration:
+  ci:
+    name: continuous-integration
     runs-on: ubuntu-latest
 
     steps:
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -18,7 +18,7 @@ on:
   workflow_dispatch:
 
 jobs:
-  build:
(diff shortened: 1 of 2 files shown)


.github/workflows/test.yml
```diff
@@ -14,7 +18,7 @@ on:
   workflow_dispatch:
 
 jobs:
-  build:
+  continuous-integration:
```
SotosTsepe: I'd highly recommend shortening the definition here, e.g. CI. Alternatively, you can use a name for the job, so it's more readable in the UI.

ValentinFutterer (approved):
Approved 👍

.github/workflows/test.yml
ValentinFutterer: Done

PR #523, opened 2026-03-30
Multitenancy: Make landing page multitenancy safe

### LINKED ISSUE #520: Multitenancy: Landing page needed
**Describe the bug**
In multitenancy mode, we have the problem of the landing page not being customisable - no-one is logged in when they hit the landing page, so we have no way of telling to which institution they belong. This means we probably need a special DAMAP cloud landing page that only triggers when multitenant mode is active. It should tell users this is DAMAP cloud you will be transferred to your instance after logging in, here is the login button or something of this nature. Ask Christiane / Tomasz how they would like it (but try to keep the amount of time needed to a minimum). 

**Is this a regression?**
No.

**To Reproduce**
Just start in multitenancy and customise.

**Expected behavior**
In multitenant mode, admins should not be able to customise landing page images. In multitenant mode, users should see a special cloud landing page or no landing page (which is the other alternative). You can ask Tomasz / Christiane what they prefer. We can also always add one later if we go with no landing page option. If you want to see how to do no landing page, check this branch https://github.com/damap-org/damap-frontend/tree/release-candidate-v5.0.0-rc1.

This PR https://github.com/damap-org/damap-backend/pull/475 will provide a config option `multitenancyEnabled` which can tell you if multitenancy is active or not. If the PR is not yet merged, you can just mock it  to test.

**Environment**

- Damap version 4.7.0

ValentinFutterer:
## Description

Bugfix

#### What does this PR do?

Made basecolor the normal blue DAMAP used before - this way, the landing page looks like it has until now in multitenancy mode. I also made the landing page images not configurable in multitenancy mode by adding a locked variable and using the backend multitenancy config variable.  This way, admins dont get the wrong idea that they are able to change the landing page.

closes GH-520

### DIFF
--- a/apps/damap-frontend/src/app/services/color-theme.service.ts
+++ b/apps/damap-frontend/src/app/services/color-theme.service.ts
@@ -69,9 +69,11 @@ export class ColorThemeService {
   private applyMaterial3Tokens(colors: Colors, exactMode: boolean): void {
     // 1. Generate Base Palette & Scheme
     const palette = CorePalette.fromColors({
-      primary: argbFromHex(colors.primary || '#6200ee'),
+      primary: argbFromHex(colors.primary || '#006699'),
       secondary: colors.secondary ? argbFromHex(colors.secondary) : undefined,
-      tertiary: colors.tertiary ? argbFromHex(colors.tertiary) : undefined,
+      tertiary: colors.tertiary
+        ? argbFromHex(colors.tertiary || '#373737')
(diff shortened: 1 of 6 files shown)


DanceaVlad (approved):
- Colors look good
- Mocked the multi-tenancy mode by hardcoding `config.mulitenancyEnabled` to true and the grid filters out images correctly and looks visually consistent

Looks good to me!

PR #525, opened 2026-04-07
Vd/fix config loading on multitenancy

### LINKED ISSUE #518: Multitenancy: Custom colors and images dont load without refresh
**Describe the bug**
In normal mode, custom colors and images are loaded with the config without a problem. In multitenancy, on first config load, the system doesnt yet know to which tenant you belong, so it loads the configured from the bootstrap db (which are none). Then it does the login flow and continues with the bootstrap values.

**Is this a regression?**
No.

**To Reproduce**

Start system in dev as multitenancy mode like this: `mvn quarkus:dev "-Dquarkus.profile=dev, multitenant"`
To simulate different tenants, you can either generate more users in the keycloak and give them an affiliation field or just directly manipulate the getAffiliation method in the code. You also need to generate an extra db per tenant and fill out the tenants.yml file. These processes are described in CUSTOMISING.md.

Then just customise the application log out and log in to observe the problem.

You may need to disable `DamapSecurityAugmentor` in the Backend since there is a bug.

**Expected behavior**
First off, customisations should immediately load after login. There could be an additional problem with images, since they might not be replaced by loading the config twice. Maybe solvable by loading customization in a sexcond wave with a different config?

**Environment**
- Damap version 4.7.0

ValentinFutterer:
## Description

Bugfix

#### What does this PR do?

Fixes an issue in multitenancy mode where configs are loaded before user logs in - making it impossible to load tenant specific configs. The solution employed in this PR uses the most anive solution of just reloading configs after the user has logged in, replacing the defaults with tenant specific configurations like images or urls.

closes GH-518

### DIFF
--- a/apps/damap-frontend/src/app/services/config.service.spec.ts
+++ b/apps/damap-frontend/src/app/services/config.service.spec.ts
@@ -36,9 +36,12 @@ describe('ConfigService', () => {
     env: 'test-env',
     appTitle: 'Test App Title',
     personSearchServiceConfigs: [],
+    projectSearchServiceConfig: null,
     fitsServiceAvailable: false,
     livePreviewAvailable: true,
     ethicalReportEnabled: true,
+    multitenancyEnabled: false,
+    templates: [],
     images: [
       {
         id: 1,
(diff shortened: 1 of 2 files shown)


ValentinFutterer (commented):
Looks good, I also added handling of app title and rudimentary error handling after the second config call

PR #534, opened 2026-04-15
Remove default banner flickering on startup

DanceaVlad:
## Description

Bugfix

#### What does this PR do?

On startup, the main page would initialize an empty default banner. When the backend config would come back, saying that there is no banner present, it would disappear. This caused heavy flickering on every reload.
Now the banner is initialized as invisible and only turns visible WHEN / IF the backend config says that there is a banner present.


ValentinFutterer (approved):
Works well, its better to have flickering of something that should be there instead of something that disappears again.

PR #536, opened 2026-04-15
translation: implement new translation key system and translation UI, replace all old keys

GeoffreyKarnbach:
## Description

Replace old key structure by new key structure in HTML and TS files using python script
Fix DMP module to make usage of new translation system
Remove all old I18n translation JSON files, load translations dynamically from DB
Implement translation management dashboard in admin pannel.

This is the script that was used to replace the old keys (some had to be done by hand):
[golden_master.csv](https://github.com/user-attachments/files/27008733/golden_master.csv)
[translation_replacer.py](https://github.com/user-attachments/files/27008743/translation_replacer.py)
[tsconfig.tools.json](https://github.com/user-attachments/files/27008747/tsconfig.tools.json)


ValentinFutterer (approved):
Looks like it works

PR #540, opened 2026-04-27
feat: add instance configuration (public availability + consent toggle) with lock page and route guard

GeoffreyKarnbach:
## Description

Feature

#### What does this PR do?

Adds frontend support for instance-level configuration.

This PR introduces:
- an instance locked page for non-admin users
- route guarding based on the `publicAvailable` config flag
- admin toggles for public availability and consent form visibility

When an instance is not public, non-admin users are redirected to the locked page. Admin users can still access the application and manage the instance configuration.

#### Dependencies

Requires backend support for instance config:
- `GET /api/admin/instance-config`
- `PUT /api/admin/instance-config`
- config response including `publicAvailable` and `consentFormEnabled`

### DIFF
--- a/apps/damap-frontend/src/app/app.routes.ts
+++ b/apps/damap-frontend/src/app/app.routes.ts
@@ -19,7 +19,7 @@ export const APP_ROUTES: Routes = [
   },
   {
     path: 'instance-locked',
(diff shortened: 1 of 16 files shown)


ValentinFutterer (approved):
👍

PR #546, opened 2026-05-05
Show custom values on deactivated lanuguages. Persist filters between reloads.

DanceaVlad:
Show custom values while on a deactivated language.
Persist filters between admin translations dashboard reloads.


ValentinFutterer (approved):
I removed the persisting of filters between reloads since there isnt a lot of time to review before the big release and getting resetted to the top of the page / having to wait 1sec everytime you make a change is really annoying.

The upside also isnt that big, since it only affects the few translations visible on the actual translations page.

Maybe the library offers a way to reload translations, without having to reload the whole page?

PR #551, opened 2026-05-27
translation: only allow for ISO 639-1 language codes for application translations and internal storage translations

### LINKED ISSUE #549: Custom Storage Translations Language Codes
**Describe the bug**
Storages allow for translations of storage text by choosing predefined language codes:

Image

But the new translation feature has a fixed english code and allows for free language code input for new translations. Both systems are not compatible. We should allow for free input of language codes to match the translation feature. Maybe with some options of already existing language codes to make it easier fro admins to match the correct one.

**Is this a regression?**
Yes.

**Expected behavior**
- [ ] Allow for inputing of language codes exactly like in translations
- [ ] System to make sure that admins select matching language codes (something like offering options, warning, info text, etc...)
- [ ] Old storage translations should be migrated gracefully

### LINKED ISSUE #550: Limit language codes of custom translations to ISO 639 2 letter codes
**Description**
Currently any input is allowed, we could use a predifend json with all language codes + their full language name to create an easy to use autocomplete dropdown. Advantage would be that we can then guarantee that the browser language detection feature will work everytime.

Alternatively, we can write some info explaining that admins should stick to ISO 639, but its not enforced.

GeoffreyKarnbach:
## Description

Feature / Bugfix

#### What does this PR do?

This PR improves language-code handling for app translations and internal storage translations.

Previously, language selection was either a free-text input or a fixed dropdown limited to `eng` and `deu`. This made it possible to create invalid app translation languages, while internal storage translations could not use newly added app translation languages.

The PR changes this by:

- adding a shared ISO 639-1 language-code list in the frontend
- replacing the app translation language input with an autocomplete
- replacing internal storage translation language dropdowns with the same autocomplete
- displaying language options as `English (EN)`, `German (DE)`, etc.
- storing only normalized two-letter language codes such as `en`, `de`, `fr`
- validating language codes in the frontend before submitting
- preventing duplicate app translation languages in the add-language dialog
- updating internal storage title resolution to use the currently selected frontend language instead of hardcoded `eng`
- adding fallback logic for old `eng` / `deu` values while the migration/backend changes are applied

This makes app translations and internal storage translations use the same language-code format and avoids mismatches between the two systems.

#### Breaking changes

Potentially breaking for existing internal storage translations that still use legacy three-letter codes (`eng`, `deu`).

A backend/database migration is expected to migrate:

- `eng` → `en`
- `deu` → `de`

The frontend includes temporary normalization/fallback logic for legacy values.

#### Dependencies

Requires corresponding backend changes to:

- validate app translation and internal storage translation language codes against the same ISO 639-1 format
- migrate existing internal storage translation values from `eng` / `deu` to `en` / `de`
- remove or adjust the old database-level `e_language` restriction if it still only allows `eng` and `deu`

closes GH-549
closes GH-550


ValentinFutterer (approved):
There was a lot of unnecessary code and code duplication. I refactored the language code picker into a reusable component and removed all the normalization logic, as it was completely unnecessary.

PR #552, opened 2026-05-27
Run evaluation benchmarks on DMP

DanceaVlad:
#### What does this PR do?

Upon saving a DMP, the summary step shows an evaluation service benchmarks dropdown and an  "Acess" button which 
sends runs the selected benchmark on it.

#### Code review focus

Error handling for when the evaluation service is down / not responding. UI

#### Dependencies

[Backend PR 514](https://github.com/damap-org/damap-backend/pull/514)


libs/damap/src/lib/components/dmp/summary/summary.component.css
```diff
@@ -19,3 +19,194 @@ td a {
...
+.eval-head {
+  display: flex;
+  flex-direction: column;
+  gap: 0.15rem;
+}
+
+.eval-title {
+  font-size: 1rem;
+  font-weight: 600;
+  color: #1a1a1a;
```
ValentinFutterer: Please use colors from the color system and dont hardcode them

libs/damap/src/lib/components/dmp/summary/summary.component.css
```diff
@@ -19,3 +19,194 @@ td a {
...
+}
+
+.eval-subtitle {
+  font-size: 0.875rem;
```
ValentinFutterer: Same for fonts

libs/damap/src/lib/components/dmp/summary/summary.component.ts
```diff
@@ -19,7 +34,7 @@ import { MatStepper } from '@angular/material/stepper';
   standalone: false,
 })
 export class SummaryComponent implements OnInit {
-  constructor(public store: Store<AppState>) {}
+  private readonly DEFAULT_BENCHMARK_ID = '69ef5cdfcde500798dbd1af8'; // FWF Benchmark
```
ValentinFutterer, line 37: There is no guarantee that other services will have this exact benchmark. FWF is a purely austrian thing, what if the polish use their own service and they dont support it.

ValentinFutterer (approved):
Looks good, but the code needs to be adapted to DAMAPs coding practices and the results in the UI need to be pretified more. We will do that in the next release.

PR #553, opened 2026-06-03
bugfix: select correct language when browser language is not available in application translations

GeoffreyKarnbach:
## Description

Bugfix

#### What does this PR do?

This PR fixes language initialization on the landing page.

Previously, the frontend selected either the browser language or English based only on a hardcoded `en|de` check. This could still select a language for which no backend translations exist, resulting in empty translations being loaded.

Now the frontend first checks whether translations exist for the preferred language from `localStorage` or the browser. If no translations are returned, it falls back to English and stores the resolved language in `localStorage`.

### DIFF
--- a/apps/damap-frontend/src/app/components/landing-page/landing-page.component.ts
+++ b/apps/damap-frontend/src/app/components/landing-page/landing-page.component.ts
@@ -49,9 +49,9 @@ export class LandingPageComponent implements OnInit {
   }
 
   ngOnInit(): void {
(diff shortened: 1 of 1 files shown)


ValentinFutterer (approved):
It fixes the original bug where a German browser wouldnt show a translation if german wasnt configured as a translation. But if english as a language is deactivated, this wont work since the english fallback language cant be fetched.

We should fix this in a future PR, for now this is fine.

PR #561, opened 2026-06-10
Migrate to Angular 22

vlad-dancea:
Prepare for a massive dump of information on a commit basis:

# Commit 1: migrate control flow
- ran: `ng generate @angular/core:control-flow`
- `*ngIf` -> `@if` worked very well, found no bugs
- `*ngFor` -> `@for` is the opposite:
- `@for` requires a **tracking function**. This is a performance optimization: it tells the for loop which field of the object provides a unique key, thus Angular knows to only change one entry of the array instead of re-rendering all of it.
- the migration solves this by tracking entire objects (very bad). Many situations are solved by simply using `track object.id` (where an id is available of course). Other situations are not so easily solved. Some ids are not fetched instantly from the backend and you get warnings in the console that multiple entities have a null key by which they are differentiated, thus Angular re-reders the entire list.
- from my observations until now I can tell that optional values do not allow you to specify a field by which to track

# Commit 2: fix some @for tracking functions
- some of the situations previously explained were easy to solve with concatenations of non unique fields like `file.name + file.lastUpdated`
- other ones remain and are too complicated to solve. Here is a summary from Codex:
```
Backend/DTO Objects Without IDs
  - libs/damap/src/lib/components/gdpr/gdpr.component.html:15
      - item of gdpr$ | async; track item
      - Options: track item.entity if entity is unique; add an id in the GDPR DTO; use track $index if this is static display-only output.

  - libs/damap/src/lib/components/gdpr/gdpr.component.html:22
      - entry of item.entries; track entry
      - Options: normalize entries to objects with ids/keys; track a deterministic composite from entry contents; use $index if order is fixed and entries are display-only.

  - libs/damap/src/lib/components/dmp/specify-data/dataset-information/dataset-information.component.html:64
      - resource of dataset.technicalResources; track resource
      - Options: add id or referenceHash to TechnicalResource; use resource.name only if unique; use name + description; $index is acceptable for read-only display.

  - libs/damap/src/lib/widgets/license-wizard/license-wizard-dialog.html:30
      - licenses of dataLicenses; track licenses
      - Options: track a composite of contained license ids, e.g. licenses.map(l => l.id).join('|'); add a named group id; leaving object tracking is low risk because DataLicenses is static module data.
```

# Commit 3: migrate standalone
- ran `ng g @angular/core:standalone` and selected `Convert all components, directives and pipes to standalone`
- app builds and runs. Translations are being fetched but are not displayed on the screen, so keys appear instead.

# Commit 4: fix translations
- the migration pulled in the `MockTranslatePipe` instead of the real one. The fix was to remove the import from all non-test files.

# Commit 5: migrate standalone part 2
- ran `ng g @angular/core:standalone` and selected `Remove unnecessary NgModule classes`
- everything worked perfectly, no issues or errors

# Commit 6: migrate standalone part 3
- ran `ng g @angular/core:standalone` and selected `Bootstrap the project using standalone APIs`
- everything worked perfectly, no issues or errors

# Commit 7: fix all `@for` tracking functions
- the `@for` loops that were left are either view-only or properly tracked by just identity or index

# Commit 8: fix tests (see issue)
- some tests were failing due to the migrations, those were easy fixes
- some tests in `/libs` were failing even before the migrations, they are not run anywhere, I left those as is (do not compile)

# Commit 9: remove unused `@angular-builder/custom-webpack`
- remove  `@angular-builder/custom-webpack` from `package.json` and it's configuration
- `angular.json` uses the stock browser builder, the custom one is never used

# Commit 10:  remove dead NgRx login code
- removed login NgRx code, it was never used / called, we rely on `auth.service.ts` now

# Commit 11: migrate internal-storage from NgRx to a signal store
- remove state files for internal-storage
- add `internal-storage-api.ts` and `internal-storage-store.ts`
- the store uses the new `Resource` API for reactive fetching

 # Commit 12: migrate repository from NgRx to a signal store
 - remove state files for repositories
  - add `repository.api.ts and `repository.store.ts`

# Commit 13: add AI configurations for Codex and Claude Code
- add `AGENTS.md` and `CLAUDE.md` with the official instructions from the Angular team for Angular 22
- add angular-developer skill for both Codex and Claude code, the official skill from the Angular team for Angular 22
- add the Angular MCP server for both Codex and Claude code, the offical MCP server from the Angular team for Angular 22

# Commit 14: migrate dmp form and dmp from NgRx to a signal store
- remove state files for dmp and dmp-form
- add `dmp/dmp.api.ts` and `dmp/dmp-form.store.ts`

# Commit 15: remove ngrx references from tests
- fixed compilation of tests for upcoming removal of ngrx, tests still fail

# Commit 16: remove NgRx dependencies
- remove all ngrx dependencies from `package.json` and `package-lock.json`

# Commit 17: migrate Angular to v20
- migrate angular, cdk, cli, material, eslint to v20
- migrate CSS tokens to Material v20

# Commit 18: format files with eslint 20

# Commit 19: update angular-oauth2-oidc to v20

# Commit 20: update Angular to v21
- migrate angular, cdl, cli, material to v21

# Commit 21: migrate ngx-translate to v18
- upgraded app integration from `TranslateModule.forRoot()` to the v18 provider API with `provideTranslateService()` and `provideTranslateLoader()`
- replaced removed `TranslateModule` imports with standalone `TranslatePipe` and `TranslateDirective` imports
- Updated changed `TranslateService` APIs:
  - `currentLang` snapshot usage -> `getCurrentLang()`
  - `defaultLang` -> `getFallbackLang()`
  - `langs` -> `getLangs()`

# Commit 22: TEMPORARY: remove eslint v21
- temporary uninstall of eslint to avoid running v22 migration with `--force`

# Commit 23: migrate Angular to v22, install eslint
- IMPORTANT: Angular 22 requires Node v22 or above
- migrated Angular, CLI, CDK, Material, Material Moment Adapter, build tooling, compiler tooling, ng-packagr, and TypeScript to Angular 22-compatible versions
- optional Karma-to-Vitest migration was also selected for the `damap-frontend` app
  - `damap-frontend` now uses the Angular unit-test builder with Vitest
  - the `damap` library still uses the existing Karma setup
  - the migration reported that the old custom Karma config may still need manual follow-up
- Angular v22 added `ChangeDetectionStrategy.Eager` to existing components to preserve the previous default change detection behavior
- Angular v22 wrapped several optional chaining template expressions with `$safeNavigationMigration()` to preserve pre-v22 behavior
    - removed `extendedDiagnostics` blocks that were incompatible with `strictTemplates: false`
- added `ignoreDeprecations: "6.0"` for existing TypeScript 6 deprecation warnings
- explicitly set `strict: false` because this codebase has strict mode intentionally deferred

# Commit 24: migrate eslint to flat config
- migrated from legacy `.eslintrc.json` / `.eslintignore` config files to ESLint 9 flat config
- added a root `eslint.config.js`
- added `angular-eslint` and `typescript-eslint` flat-config helper packages

# Commit 25: migrate inputs to signals
- ran the official Angular migration: `ng generate @angular/core:signal-input-migration`

# Commit 26: migrate outputs to signals
- ran the official Angular migration: `ng generate @angular/core:output-migration`

# Commit 27: migrate components to self-closing tags
- ran the offiial Angular migration: `ng generate @angular/core:self-closing-tag`

# Commit 28: migrate ngclass and ngstyle to [style] and [class]
- ran the official Angular migrations:
  - `ng generate @angular/core:ngclass-to-class`
  - `ng generate @angular/core:ngstyle-to-style` 

# Commit 29: refactor: remove remaining runtime NgModules
- remove all runtime NgModules and make all components standalone.

# Commit 30: fix: render info-card step numbers as text instead of icon
- Angular Material v22 was causing the text inside <mat-icon> to appear very distorted
- add differentiation between icon and text inside the bubble

# Commit 31: add documentation for futurea Angular ChangeDetection migration
- wrote documentation for how ChangeDetection and zone.js migration should be done in the future

# Commit 32: remove e2e tests
- e2e tests were very outdated and very prone to breaking due to overfitted css selectors

# Commit 32: remove spec tests
- spec tests were also either outdated, broken, or "should create" smoke tests

# Commit 33: BIG: remove project boundary between shell and core
- TU Graz no longer relies on the separation of damap-shell and damap-core as 2 different Angular projects
- merged the 2 in a single project, kept folder based separation for organizational purposes

# Commit 34: add new linting and formatting setup
- merged all eslint.config files into 5 files in the project root.
- Typescript 6 is strict by default, but keeping that on results in hundreds of linting errors.
- My proposed convention:
  - keep strict off for the whole project
  - enforce strict for new files: **/*.spec.ts, **/*.store.ts, **/*.api.ts
  - whenever migrating a component to new signal-store + api combo, add it to the `tsconfig.strict.ts`
  - add `npm run typecheck:strict` to CI

# Commit 35: add vitest configuration and sample test
- add `vitest.config.ts`
- force strict typechecking on spec files

# Commit 36: make store and api files adhere to strict TS

# Commit 37: add playwright configuration and sample tests
- add `playwright.config.ts`
- run e2e tests either headless or with browser.

# Commit 38: add documentation about testing and linting
- add `DEVELOPMENT.md` file for documenting linting, testing and npm commands

# Commit 39: add new ci commands to GH workflow


lpandath:
@DanceaVlad 
This part: "migrate dmp form and dmp from NgRx to a signal store": 
the DMP form has a lot of moving parts (multi-step navigation, autosave, draft, versioning). Have you tested the full flow end to end? If you havealready done a manual test pass, could you describe what you tried? That would help me downstream when we update our own instances.

On the PR description's list of unresolved @for tracking spots:

Here description still mentions a few unresolved @for tracking issues for gdpr.component.html, dataset-information.component.html, license-wizard-dialog.html. One of your later commits says "fix all @for tracking functions" did those get covered there? or is the description out of date? Just want to make sure nothing was forgotten.

lpandath:
@DanceaVlad  One thing that i mentioned in the meeting too, suggestion about structuring the migration:
if  possible to keep a short migration doc in the repo (something like migrations/v22.md or a folder with one file per major step) describing the changes you made and why...

apps/damap-core/src/app/data-access/repository.store.ts
```diff
@@ -0,0 +1,102 @@
...
+
+import { LoadingState } from '../domain/enum/loading-state.enum';
+import { RepositoryDetails } from '../domain/repository-details';
+import { RepositoryApi, RepositoryFilter } from './repository.api';
+
+function hasActiveFilters(filter: RepositoryFilter): boolean {
+  return Object.keys(filter).some((key) => filter[key]?.length > 0);
+}
+
+@Injectable()
```
lpandath, line 13: RepositoryStore is @Injectable() while InternalStorageStore is @Injectable({ providedIn: 'root' }). Is the scope intentional for repositories example,  so filter state resets per page? If so, worth a comment explaining it, if not, probably should match.

apps/damap-core/src/app/data-access/dmp.api.ts
```diff
@@ -0,0 +1,85 @@
...
+  createDmp(dmp: Dmp): Observable<Dmp> {
+    return this.http.post<Dmp>(this.dmpUrl, dmp, this.jsonOptions).pipe(retry(3));
+  }
+
+  updateDmp(dmp: Dmp): Observable<Dmp> {
+    return this.http.put<Dmp>(`${this.dmpUrl}/${dmp.id}`, dmp, this.jsonOptions).pipe(retry(3));
+  }
+
+  saveDmpVersion(version: DmpVersionDraft): Observable<Version> {
+    return this.http.put<Version>(this.versionUrl, version, this.jsonOptions).pipe(retry(3));
```
lpandath: POST/PUT calls is unsafe somehow. If the request reaches the server and creates/updates the DMP but the response gets lost where retrying will create duplicate DMPs on POST or potentially overwrite intermediate edits on PUT. POST/PUT need either idmp keys a client-generated UUID the backend uses to dedupe or no retry at all. GET retries are fine but mutation retries usually aren't no?!

lpandath: In general the files in the three *.api.ts files use very different conventions internal-storage.api.ts is just a baseUrl, repository.api.ts is URLs + helpers, but dmp.api.ts carries full HTTP methods. Is the intent for *.api.ts to be URL builders only? or are the other two going to grow methods too?

apps/damap-core/src/app/data-access/dmp.store.ts
```diff
@@ -0,0 +1,175 @@
...
+    this.savingDmpState.set(true);
+    return this.api.createDmp(dmp).pipe(
+      tap((savedDmp) => {
+        this.formService.mapDmpToForm(savedDmp);
+        this.feedbackService.success('dmp.success.save');
+        this.loadDmps(false);
+      }),
+      catchError((error: Error) => {
+        this.feedbackService.error('dmp.save.error', error);
+        return EMPTY;
```
lpandath: catchError returning EMPTY swallows errors callers can't know if save failed.

apps/damap-core/src/app/data-access/dmp.store.ts
```diff
@@ -0,0 +1,175 @@
...
+    this.dmpsResource.update((dmps) => dmps.filter((dmp) => dmp.id !== id));
+  }
+
+  createDmp(dmp: Dmp): Observable<Dmp> {
+    this.savingDmpState.set(true);
+    return this.api.createDmp(dmp).pipe(
+      tap((savedDmp) => {
+        this.formService.mapDmpToForm(savedDmp);
+        this.feedbackService.success('dmp.success.save');
+        this.loadDmps(false);
```
lpandath, line 67: this doesn't refresh the list no? stale data after edit. Also no loadDmps call

apps/damap-core/src/app/data-access/dmp-form.store.ts
```diff
@@ -0,0 +1,103 @@
...
+    return a === b;
+  }
+
+  const aKeys = Object.keys(a);
+  const bKeys = Object.keys(b);
+  if (aKeys.length !== bKeys.length) {
+    return false;
+  }
+
+  return aKeys.every((key) => equals(a[key], b[key]));
```
lpandath: Object.keys(new Date()) returns [] because Date has no own enumerable properties. So if you compare new Date('2026-01-01') with new Date('2026-12-31'), both keys arrays are length 0?

apps/damap-core/src/app/data-access/dmp.api.ts
```diff
@@ -0,0 +1,85 @@
...
+  }
+
+  private downloadFile(response: HttpResponse<Blob>): void {
+    if (!response.body) {
+      throw new Error('Cannot download an empty DMP export response.');
+    }
+
+    const contentDisposition = response.headers.get('content-disposition');
+    const filename = contentDisposition
+      ? contentDisposition.substring(contentDisposition.lastIndexOf('filename=') + 9)
```
lpandath, line 75: If DMPs with titles containing spaces or special characters end up in content-disposition, the download filename will be wrong

apps/damap-core/src/app/components/dmp/dmp.component.ts
```diff
@@ -33,34 +26,89 @@ import { InfoLabelService } from '../../services/infoLabel.service';
...
+    LicensesComponent,
+    RepoComponent,
+    ReuseComponent,
+    CostsComponent,
+    SummaryComponent,
+    DmpActionsComponent,
+    TranslatePipe,
+  ],
 })
 export class DmpComponent implements OnInit, OnDestroy {
```
lpandath, line 98: Comment on the hole file. Here the form data lives in both dmpForm and formStore.dmp() (signal store), kept in manual sync via valueChanges.subscribe to  formStore.formDiff + an effect that reads back. I find it a bit the trickiest spot in the migration and probably why setTimeout in ngOnInit, several manual cdr.detectChanges() calls, and the auto-added ChangeDetectionStrategy.Eager are all needed here. Worth deciding whether to commit fully to signals (drop the reactive form) or the other way around?

apps/damap-core/src/app/data-access/repository.store.ts
ValentinFutterer: I think its only `@Injectable() ` because in the future we want to have two stores, one for step 8 and one for admin. I went through all other `@Injectable() ` and added the root provider when necessary, most relevant are the gaurds.

apps/damap-core/src/app/components/admin/edit-repositories-page/edit-repositories-page.component.ts
ValentinFutterer: Remove this key http.error.repositories.all

apps/damap-core/src/app/services/backend.service.ts
```diff
@@ -244,9 +171,7 @@ export class BackendService {
   getInternalStorages(): Observable<InternalStorage[]> {
     const langCode = 'eng'; // TODO: Replace with template lang in the future
     return this.http
-      .get<
-        InternalStorage[]
-      >(`${this.backendUrl}storages?languageCode=${langCode}`)
+      .get<InternalStorage[]>(`${this.backendUrl}storages?languageCode=${langCode}`)
       .pipe(retry(3), catchError(this.handleError('http.error.storages')));
```
ValentinFutterer: Remove 'http.error.storages' as this emthod can mbe removed - check first if the message isnt needed at another place though!

apps/damap-core/src/app/data-access/dmp.api.ts
ValentinFutterer: Good point with the retries - though they werent added during the upgrade to angular 22, they were already present in the old backend service methods. I have removed them.

AGENTS.md
```diff
@@ -0,0 +1,136 @@
...
+- Angular v20 stabilized incremental hydration and route-level render modes.
+- Angular v20.2 stabilized zoneless Angular.
+- Angular v21 made zoneless the default for new apps and made Vitest the primary test runner for new projects.
+- Angular v22 continues the modern defaults: standalone, signals, native control flow, deferrable views, zoneless-compatible components, Signal Forms, resources, `httpResource`, CLI AI config, and MCP-assisted tooling.
+
+When these instructions conflict with current Angular MCP documentation for the installed framework version, trust Angular MCP and update the instructions if the change is durable.
+
+# Damap Frontend
+
+// TODO: Add project guidelines and descriptions here
```
ValentinFutterer, line 136: Add to PR template that we shouldalways try to write new api - store patterns in place of backend.service if possible and we should remove the eager updating whenever we see it.

apps/damap-core/src/app/data-access/dmp.api.ts
ValentinFutterer: The point for the *.api.ts files is to be an interface with the backend - they are supposed to only care about fetching data and no state changes. The store handles how the fetched data affects the surroundings. I will write some docs to explain this pattern.

apps/damap-core/src/app/data-access/dmp.store.ts
ValentinFutterer: Fixed it by implementing a global error handler.

apps/damap-core/src/app/data-access/dmp.store.ts
ValentinFutterer: What list?  The DMP list? `loadDmps` is called right.

apps/damap-core/src/app/components/dmp/dmp.component.ts
ValentinFutterer: I removed the form store, the manual detectChanges calls and the timeout. Seems to work fine without it. But the `ChangeDetectionStrategy.Eager ` is apparently still needed. I will keep it in, since the whole project is already full with them and we can fix it as we go (will write docs)

apps/damap-core/src/app/data-access/dmp-form.store.ts
ValentinFutterer: I removed the form store.

apps/damap-core/src/app/data-access/dmp.api.ts
ValentinFutterer: Has this been a problem before?
Because the old implementation worked exactly the same: https://github.com/damap-org/damap-frontend/pull/561/changes#diff-75c8e689d8f244eed93c4939bff1ecdd0a2f92fb8b8ee0981e1dfbc8689c53d9L60-L65

apps/damap-core/src/app/data-access/repository.store.ts
lpandath: The scope itself is fine, but right now the store is provided twice and since the table is a child of the repo step, it gets its own second instance. Each instance creates its `httpResource`s eagerly, so opening step 8 already fetches the full `/repositories` list nobody uses and opening the table fetches `/repositories/recommended` a second time. Also `getRepositoryDetails` in `repo.component.ts:94` writes into the parent instance, which no template reads?!

apps/damap-core/src/app/data-access/dmp.api.ts
lpandath: Can the docs you mentioned land in this PR? The project section in AGENTS.md is still a TODO at line 136?! and would be the natural place

apps/damap-core/src/app/data-access/dmp.store.ts
lpandath: In the `catchError` pipes that is fine but the resource error effects call it bare, without await or catch

apps/damap-core/src/app/data-access/dmp.api.ts
lpandath: yes is not a blocker for this PR, just two notes for a follow-up: the same parser now lives twice in backend service and here, and neither strips quotes or handles `filename*=`.  we could eventually open a separate issue

ValentinFutterer: Can you handle that, pretty please ? :)

apps/damap-core/src/app/data-access/dmp.store.ts
ValentinFutterer: Yeah there is no reason to catch it here. The resource hides the HTTPClient, so its impossible to use the `catchError` from rxjs. The returned value is just ignored here, so the async has no ill effect.

apps/damap-core/src/app/data-access/dmp.api.ts
ValentinFutterer: I am going to remove the agents part, since they are unused, need to be maintained and have no licenses attached. Docs will be in seperate files.

apps/damap-core/src/app/data-access/repository.store.ts
ValentinFutterer: Yes thats on purpose. the repo store maintains a state of e.g. current filters. I think that having one global instance would mean that an admin will see some search behaviour being carried between the admin panel and step 8. To avoid this, its injected twice. This is only ever going to affect admins, so a very small percentage of the traffic.

The eager fetching behaviour has always been there - its not a big problem as its cached in the backend anyway.

What do you mean with writes to a parent? Did you flag the wrong codeline by mistake?

PR #570, opened 2026-07-09
Layout / Admin: Configurable accessibility statement link in sidebar footer

### LINKED ISSUE #569: Add configurable Accessibility Statement link to sidebar footer
**Description**
TU Graz has had an accessibility statement link in the DAMAP footer since v4.x, following the same pattern as other TU Graz services (example: https://repository.tugraz.at/records/s2nwf-crj56).

We'd like to carry this link over into v5, but driven by instance_config rather than baked into the fork so the URL is admin-editable and each deployment can configure their own statement page or leave it empty if they don't need one.

- New translation key sidebar.footer.accessibility-statement.link for the label.
- Read footer_accessibility_url from instance_config 
- If set → render the footer link. If null/empty.

**Describe alternatives you've considered**

**Additional context**
[Add any other context or screenshots about the feature request here.](https://github.com/damap-org/damap-backend/issues/523)

_If this issue requests multiple changes consider using checkboxes or lists._

lpandath:
## Description

Adds an optional fourth link "Accessibility Statement" to the sidebar footer, plus an admin editor to configure the URL from `instance_config`. The link renders only when the URL is set.

#### What does this PR do?

Storing the URL in `instance_config` keeps this out of per-fork patches each institution sets their own statement URL from the Admin page.

#### Code review focus

Backend companion PR: damap-org/damap-backend#524

### Checks

- [ ] Summary updated
- [ ] Version view updated
- [ ] Documentation added
- [ ] Tests added
- [ ] E2e tests created

closes GH-

### DIFF
--- a/apps/damap-frontend/src/assets/i18n/translations.json
+++ b/apps/damap-frontend/src/assets/i18n/translations.json
@@ -20,6 +20,8 @@
     "sidebar.footer.version": "Version",
     "sidebar.footer.data-protection-declaration": "Data Protection Declaration",
     "sidebar.footer.data-protection-declaration.link": "#",
+    "sidebar.footer.accessibility-statement": "Accessibility Statement",
+    "sidebar.footer.accessibility-statement.label": "Accessibility Statement",
     "sidebar.footer.collapsed.hint": "Expand the menu to access legal information",
     "dashboard.home.section-intro": "Welcome to DAMAP, a service that helps you create a Data Management Plan (DMP) for your project.",
     "dashboard.cards.instructions-info": "You can create a new DMP by clicking the 'Create New DMP' button.",
--- a/apps/damap-frontend/src/assets/i18n/translations.json
+++ b/apps/damap-frontend/src/assets/i18n/translations.json
@@ -20,8 +20,6 @@
     "sidebar.footer.version": "Version",
     "sidebar.footer.data-protection-declaration": "Data Protection Declaration",
     "sidebar.footer.data-protection-declaration.link": "#",
-    "sidebar.footer.accessibility-statement": "Accessibility Statement",
-    "sidebar.footer.accessibility-statement.label": "Accessibility Statement",
     "sidebar.footer.collapsed.hint": "Expand the menu to access legal information",
     "dashboard.home.section-intro": "Welcome to DAMAP, a service that helps you create a Data Management Plan (DMP) for your project.",
     "dashboard.cards.instructions-info": "You can create a new DMP by clicking the 'Create New DMP' button.",
--- a/libs/damap/src/lib/components/admin/admin.component.html
+++ b/libs/damap/src/lib/components/admin/admin.component.html
@@ -163,5 +163,49 @@
         ) | translate
       }}
     </p>
+
+    <mat-slide-toggle
+      [checked]="footerAccessibilityEnabled"
+      (change)="onFooterAccessibilityEnabledToggle($event.checked)">
+      {{ "admin.instance-config.footer-accessibility-enabled" | translate }}
+    </mat-slide-toggle>
+
+    <p class="instance-config-hint">
+      {{
+        (footerAccessibilityEnabled
+          ? "admin.instance-config.footer-accessibility-enabled-on"
+          : "admin.instance-config.footer-accessibility-enabled-off"
+        ) | translate
+      }}
+    </p>
+
+    @if (footerAccessibilityEnabled) {
+      <form (ngSubmit)="saveFooterAccessibilityUrl()">
+        <app-input-wrapper
+          [label]="'admin.instance-config.footer-accessibility-url' | translate"
+          [control]="footerAccessibilityUrlControl"
+          [type]="'url'"
+          [maxLength]="2048"></app-input-wrapper>
+
+        <div class="instance-config-actions">
+          <button
+            type="submit"
+            mat-flat-button
+            color="primary"
+            [disabled]="footerAccessibilityUrlControl.invalid">
+            {{ "admin.instance-config.footer-accessibility-save" | translate }}
+          </button>
+          <button
+            type="button"
+            mat-stroked-button
+            [disabled]="!instanceConfig?.footerAccessibilityUrl"
+            (click)="removeFooterAccessibilityUrl()">
+            {{
+              "admin.instance-config.footer-accessibility-remove" | translate
+            }}
+          </button>
+        </div>
+      </form>
+    }
   </div>
 </div>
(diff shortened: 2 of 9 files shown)


apps/damap-frontend/src/assets/i18n/translations.json
```diff
@@ -20,6 +20,8 @@
     "sidebar.footer.version": "Version",
     "sidebar.footer.data-protection-declaration": "Data Protection Declaration",
     "sidebar.footer.data-protection-declaration.link": "#",
+    "sidebar.footer.accessibility-statement": "Accessibility Statement",
+    "sidebar.footer.accessibility-statement.label": "Accessibility Statement",
```
ValentinFutterer: This file isnt even up to date anymore and should have been removed. There is no need to add the translations here.

libs/damap/src/lib/components/admin/admin.component.ts
```diff
@@ -39,6 +45,15 @@ export class AdminComponent implements OnInit {
   selectedInternalStorageUrl: string;
 
   instanceConfig: InstanceConfig | null = null;
+  footerAccessibilityEnabled = false;
+  footerAccessibilityUrlControl: UntypedFormControl = new UntypedFormControl(
+    '',
+    [
+      Validators.maxLength(2048),
+      (control: AbstractControl): ValidationErrors | null =>
+        !control.value || this.isValidUrl(control.value) ? null : { url: true },
```
ValentinFutterer: Good job with the URL validation 👍

libs/damap/src/lib/components/admin/admin.component.html
```diff
@@ -163,5 +163,49 @@ <h2 translate>admin.instance-config.title</h2>
...
+          [label]="'admin.instance-config.footer-accessibility-url' | translate"
+          [control]="footerAccessibilityUrlControl"
+          [type]="'url'"
+          [maxLength]="2048"></app-input-wrapper>
+
+        <div class="instance-config-actions">
+          <button
+            type="submit"
+            mat-flat-button
+            color="primary"
```
ValentinFutterer: This was deprecated since Angular Material 3, please do not use it anymore.

libs/damap/src/lib/components/admin/admin.component.ts
```diff
@@ -317,4 +336,100 @@ export class AdminComponent implements OnInit {
...
+          config.footerAccessibilityUrl ?? '',
+        );
+        this.feedbackService.success('http.success.instance-config.update');
+      },
+      error: error => {
+        if (error.error?.message) {
+          this.feedbackService.error(error.error.message);
+        } else {
+          this.feedbackService.error(error.message);
+        }
```
ValentinFutterer: There is no reason to handle the error here, the method in the backend service already takes care of this.

libs/damap/src/lib/components/admin/admin.component.ts
```diff
@@ -317,4 +336,100 @@ export class AdminComponent implements OnInit {
...
+      return;
+    }
+
+    const trimmed = (this.footerAccessibilityUrlControl.value ?? '').trim();
+    if (!trimmed) {
+      this.feedbackService.error(
+        'admin.instance-config.footer-accessibility-url-empty',
+      );
+      return;
+    }
```
ValentinFutterer: You can remove this and just add `|| !footerAccessibilityUrlControl.value` to the save button disabled condition.

libs/damap/src/lib/components/admin/admin.component.ts
```diff
@@ -317,4 +336,100 @@ export class AdminComponent implements OnInit {
...
+        this.footerAccessibilityEnabled = false;
+        this.footerAccessibilityUrlControl.setValue('');
+        this.feedbackService.success('http.success.instance-config.update');
+      },
+      error: error => {
+        if (error.error?.message) {
+          this.feedbackService.error(error.error.message);
+        } else {
+          this.feedbackService.error(error.message);
+        }
```
ValentinFutterer: Same with the error, its handled in the service

libs/damap/src/lib/components/admin/admin.component.ts
```diff
@@ -317,4 +336,100 @@ export class AdminComponent implements OnInit {
...
+
+        this.instanceConfig = {
+          ...this.instanceConfig!,
+          footerAccessibilityUrl: previousUrl,
+        };
+        this.footerAccessibilityEnabled = !!previousUrl;
+        this.footerAccessibilityUrlControl.setValue(previousUrl);
+      },
```
ValentinFutterer: This code is completely superflous, the url is only set to `` in case of a success.

libs/damap/src/lib/components/admin/admin.component.ts
```diff
@@ -317,4 +336,100 @@ export class AdminComponent implements OnInit {
...
+
+  saveFooterAccessibilityUrl() {
+    if (!this.instanceConfig) {
+      return;
+    }
+
+    if (this.footerAccessibilityUrlControl.invalid) {
+      this.footerAccessibilityUrlControl.markAsTouched();
+      return;
+    }
```
ValentinFutterer: The button cannot be pressed when its invalid, this code can be removed.

libs/damap/src/lib/components/admin/admin.component.ts
```diff
@@ -317,4 +336,100 @@ export class AdminComponent implements OnInit {
...
+          this.feedbackService.error(error.error.message);
+        } else {
+          this.feedbackService.error(error.message);
+        }
+
+        this.instanceConfig = {
+          ...this.instanceConfig!,
+          footerAccessibilityUrl: previousUrl,
+        };
+        this.footerAccessibilityUrlControl.setValue(previousUrl);
```
ValentinFutterer: This reset is completely unnecessary, the actual value is only touched in the case of success.

libs/damap/src/lib/components/admin/admin.component.ts
lpandath: you are right for instance config reset and the setValue, i  removed, those are only touched on success. I kept one line in the error case though: onFooterAccessibilityEnabledToggle() sets footerAccessibilityEnabled = false before the request is sent, so if the update fails, the toggle would stay off while the URL is still configured the form disappears and the footer keeps showing the link until a reload. Re-syncing the flag keeps the toggle consistent with the actual config:

```
 error: () => {
   this.footerAccessibilityEnabled =
     !!this.instanceConfig?.footerAccessibilityUrl;
 },

```
I can drop it too if you'd rather keep it minimal.

ValentinFutterer: Ah true, `onFooterAccessibilityEnabledToggle` calls the remove 👍

PR #579, opened 2026-09-17
Fix(backend.service): normalize DOI formats before dataset lookup

SotosTsepe:
## Description

Bugfix

#### What does this PR do?

Dataset lookup accepted bare DOIs, DOI URLs, and `doi:` prefixes in the frontend, but sent the input unchanged to the backend. This caused lookups with either prefix to fail.

This change strips the prefix and surrounding whitespace before sending the request. It also uses HTTP query parameters to encode special characters safely.

Regression tests cover all three formats, surrounding whitespace, and query parameter encoding.
All three formats were manually verified against a local running backend.

#### Breaking changes

None.

#### Code review focus

DOI normalization in `BackendService.searchDataset` and the regression tests for validation and request parameters.

#### Dependencies

No additional backend changes required.

### Checks

- [x] Tests added
- [x] Lint and formatting checks passed
- [x] Manually verified dataset lookup with all 3 DOI formats
- [x] Tests passing


ValentinFutterer (approved):
👍

